PR #3, opened 2016-11-01
Add LDAP button and transition code

gene1wood:
Add jQuery
Dissalow forgot password
Change social button style to dynamic

### DIFF
--- a/pages/login.html
+++ b/pages/login.html
@@ -149,6 +149,7 @@
 
     var options = {
       allowForgotPassword: false,
+      allowSignUp: false,
       auth: {
         redirectUrl: config.callbackURL,
         responseType: config.callbackOnLocationHash ? 'token' : 'code',
--- a/pages/login.html
+++ b/pages/login.html
@@ -5,7 +5,36 @@
   <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
   <title>Mozilla | Log in</title>
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <style> body,html{padding:0;margin:0}.table{display:table;position:absolute;height:100%;width:100%;background:linear-gradient(rgba(255,255,255,.3),rgba(255,255,255,0));background-color:#e8ebef}.cell{display:table-cell;vertical-align:middle}.content{padding:25px 0;margin-left:auto;margin-right:auto;width:280px}.auth0-lock-header-bg {background: #D7D3C8 !important;}
+    <style>
+      body,html {
+        padding:0;
+        margin:0;
+      }
+      .table {
+        display:table;
+        position:absolute;
+        height:100%;
+        width:100%;
+        background:linear-gradient(rgba(255,255,255,.3),rgba(255,255,255,0));
+        background-color:#e8ebef;
+      }
+      .cell {
+        display:table-cell;
+        vertical-align:middle;
+      }
+      .content {
+        padding:25px 0;
+        margin-left:auto;
+        margin-right:auto;
+        width:280px;
+      }
+      .auth0-lock-header-bg {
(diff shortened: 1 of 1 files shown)


pages/login.html
```diff
@@ -5,9 +5,58 @@
   <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
   <title>Mozilla | Log in</title>
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <style>
-	body,html{padding:0;margin:0}.table{display:table;position:absolute;height:100%;width:100%;background:linear-gradient(rgba(255,255,255,.3),rgba(255,255,255,0));background-color:#e8ebef}.cell{display:table-cell;vertical-align:middle}.content{padding:25px 0;margin-left:auto;margin-right:auto;width:280px}.auth0-lock-header-bg {background: #D7D3C8 !important;}
-	.auth0-label-submit{color:#4D4E53;font-weight:600}
```
gdestuynder: Does the new commit retain the styling for the login text (auth0-label-submit)? I can't see it but maybe it's inherited somewhere

pages/login.html
```diff
@@ -42,19 +92,41 @@
...
 
     var options = {
+      allowForgotPassword: false,
```
gdestuynder: is this required when database connections are not used? 
According to https://github.com/auth0/lock/blob/master/README.md this is for database only. Does not hurt, but if we use that "just in case" maybe we want other options.
IMO not necessary though

pages/login.html
```diff
@@ -66,13 +138,21 @@
       },
       theme: {
         logo: 'data:image/svg+xml;base64,' + window.btoa(logo),
-        primaryColor: "#D7D3C8"
+        primaryColor: "#d7d3c8"
       },
       closable: false,
-      socialButtonStyle: 'small'
+      // socialButtonStyle: 'small'
```
gdestuynder: Just remove it :)

pages/login.html
```diff
@@ -66,13 +138,21 @@
...
     };
 
     var lock = new Auth0Lock(config.clientID, config.auth0Domain, options);
+    lock.on("signin ready", function() {
+      document.querySelector(".auth0-lock-form > div").insertAdjacentHTML('beforeend', '<button class="auth0-lock-ldap-button auth0-lock-ldap-big-button" type="button"><div class="auth0-lock-ldap-button-icon"></div><div class="auth0-lock-ldap-button-text">Log in with LDAP</div></button>');
```
gdestuynder: can this be in it's own function definition for clarity?

pages/login.html
gene1wood: I didn't change it so that block of css is just carried forward from master. So I'm not sure.

pages/login.html
gene1wood: My thought was that since we never want this option displayed we should [disable it](https://auth0.com/docs/libraries/lock/v10/customization#allowforgotpassword-boolean-). So yes, it's the does not hurt category just in case.

pages/login.html
gene1wood: Will do

pages/login.html
gene1wood: Ah great point, I'll add that.

pages/login.html
gene1wood: Ahh, I [see what you're talking about](https://github.com/mozilla-iam/auth0-deploy/pull/3/files#diff-d1b26f3c70508e6f65750cdc305e0ce5L10). Sure, I'll fix that.

pages/login.html
gene1wood: Will do, I'll make a variable of the function

Issue #7, opened 2016-11-03
Enable users to change their mind and navigate back to the full list of connection types


gene1wood:

gene1wood:
Fixed by #9

Issue #6, opened 2016-11-03
Enter key for passwordless email field should submit the form


gene1wood:

gene1wood:
Fixed by #9

Issue #4, opened 2016-11-01
Put login button inside LDAP expansion


gene1wood:
No need for a login button unless you choose LDAP

gene1wood:
Fixed by #9

PR #9, opened 2016-11-03
Add passwordless authentication

gene1wood:
Add a back button to navigate back to the connection selection buttons
Change UI to clear all buttons when a user chooses passwordless or LDAP


gene1wood:
* Fixes #7 
* Fixes #6 
* Fixes #4

PR #27, opened 2016-11-09
Fix social passwordless no ldap

gene1wood:
This does 2 unrelated things (sorry). It adds Modernizr to detect browser version and conditionally set the email inputs to `email` instead of `text`. It also fixes a bug in the lock with the combination of social + passwordless + no-ldap


gene1wood:
The email input fix may be present in [v10.7.0](https://github.com/auth0/lock/tree/v10.7.0) with auth0/lock#724

PR #51, opened 2016-11-15
Make cdnjs.cloudflare.com JS load explicitly HTTPS

### LINKED ISSUE #50: Missing scheme reference on pages/login.html:189
Was doing a light code review and noticed that pages/login.html does not specify a scheme.

     <script src="//cdnjs.cloudflare.com/ajax/libs/ie8/0.2.5/ie8.js"></script>

This content is available over HTTP and HTTPS, so I think we should be more explicit about that.

claudijd:
Fixes #50


claudijd:
@gene1wood r?

PR #54, opened 2016-11-17
use mutationobserver to notice/trigger dom changes, closes #29

### LINKED ISSUE #29: When lock renders, passwordless button displays early then is hidden then revealed

jeffbryner:
Here's the result of changing to use the mutation observer and some other fixes to the email login to correct timing issues.

### DIFF
--- a/pages/login.html
+++ b/pages/login.html
@@ -198,6 +198,35 @@
  <script src="https://cdn.auth0.com/js/lock/10.5/lock.min.js"></script>
  <script src="https://cdn.auth0.com/w2/auth0-7.4.min.js"></script>
  <script>
+  
+    // set up an observer to watch for auth0 nodes to trigger our customizations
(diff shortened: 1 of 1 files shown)


gene1wood:
PR #54 swaps out the method by which we wait for elements to show up in the lock. It replaces the [window.requestAnimationFrame()](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame) method with the newer [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver).

I've put together PR #72 which also solves this problem but continues to use `requestAnimationFrame` which maintains support for IE 10 (which we'd lose with `MutationObserver`)

I'm going to merge #72 and close this PR to keep our browser support as broad as possible. Thank you for pointing me to this [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) method which I'd not seen before!

PR #59, opened 2016-11-17
Merge first true production hosted locked and related files

gdestuynder:
Remember that the data in manual/* needs to be manually merged to production after this PR is merged.


gene1wood:
I've applied the manual page change to production.

PR #71, opened 2016-11-23
Prevent `enter` key from triggering form submit

### LINKED ISSUE #52: Lock remembers LDAP session when logging back in with passwordless
Steps:
- Login to https://social-ldap-pwless.testrp.security.allizom.org/ with an LDAP login
- Logout (https://social-ldap-pwless.testrp.security.allizom.org/logout) from testrp
- Login again to https://social-ldap-pwless.testrp.security.allizom.org/ but with passwordless
=> you are immediately logged in (even thus you also get an email, there is no need to click on the link)
=> you are logged in as the LDAP user instead of the passwordless user

This seems to be because there is a browser side cookie that is sent to auth0 during the login transaction, and that behaves like a "remember last login" (even thus this option is off in our lock)

This is both a UX and security issue since the users would not expect the auth0 session to be remembered (which it should not)

gene1wood:
Fixes #52

### DIFF
--- a/pages/login.html
+++ b/pages/login.html
@@ -331,8 +331,9 @@
           $(".auth0-lock-input-username input").focus();
         }
         $("button.auth0-lock-passwordless-submit").click(sendEmail);
-        $("input[name=passwordlessEmail]").keyup(function(event) {
+        $("input[name=passwordlessEmail]").keydown(function(event) {
           if (event.keyCode == 13) {
+            event.preventDefault();
             sendEmail();
             return false;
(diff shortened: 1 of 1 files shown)


tristanweir:
Any other implication from switching from keyup to keydown?

gene1wood:
Minor timing impact causing the sendemail to occur slightly earlier. An alternative approach would be to continue to use `keyup` and catch `keydown` and throw it away with `preventDefault`

PR #72, opened 2016-11-23
Defer lock customization until the submit button exists

### LINKED ISSUE #29: When lock renders, passwordless button displays early then is hidden then revealed

gene1wood:
Fixes #29


gene1wood:
As this is a significant structural change we should do thorough testing of dev after this is deployed before promoting to prod.

PR #78, opened 2016-11-28
adding application access rule for mozdefqa app

jdow:
@gdestuynder r? is this the correct way to add specific app-to-ldap-group mappings? If so, please merge and deploy to production.


gdestuynder:
also yes :)

PR #80, opened 2016-11-29
Prevent multiple "Log in by Email" submissions

### LINKED ISSUE #47: Clicking passwordless submit twice before the success/error screen results in multiple emails and success/error pages

tristanweir:
Fixes #47
Logic works for both mouse clicks and presses of the Enter key


gene1wood:
I've looked over this code and it looks fine. The `focus` then `click` thing is a little odd but it 's fine.

Issue #75, opened 2016-11-24
Cannot login using LDAP on Mozillians.org


MikkCZ:
After mozillians.org replaced Persona by LDAP and passwordless login options, I have an issue with signind via my LDAP account there. It works on l10n.mozilla.org, passwordreset.mozilla.org and login.mozilla.com, but when using my LDAP on Mozillians.org, I end up with 2-Step Verification error "Access Denied. The username you have entered cannot authenticate with Duo Security. Please contact your system administrator.".

I have already contacted #commtools asking about this. There is know bug for the passwordless auth, but with the LDAP issue I have been directed here.

MikkCZ:
To add, I am not staff, but have the LDAP account for l10n commit access etc.

gene1wood:
@MikkCZ When you say "There is know bug for the passwordless auth" do you mean "There is known bug for the passwordless auth" or "There is no bug for the passwordless auth"?

If you mean "There is known bug for the passwordless auth" can you point me to what bug you're referring to?

MikkCZ:
I have been told on #commtools on IRC, that there is a bug for it, but do not have the link, sorry. There is a workaround (login in Firefox private mode), but that works for passwordless only, not LDAP.

gene1wood:
Ok, I'll ping around in #comtools to find out if anyone knows what bug they were referring to.

And do you consistently experience this bug with passwordless auth that you mention? Can you share some of the details about what you experience.

gene1wood:
@johngian looked through IRC logs and found this as the bug that was mentioned : https://bugzilla.mozilla.org/show_bug.cgi?id=1319822

@MikkCZ as we made a production deploy of the fix mentioned in that bug this morning could you try a passwordless login (aka email login) again and see if it works?

MikkCZ:
The e-mail login works on mozillians.org now, but does not on AirMo and some people complained about it on Mozilla Reps Telegram today too. This is what I see after pasting the code for e-mail login on AirMo:
![snimek obrazovky 2016-11-28 v 20 40 27-fullpage](https://cloud.githubusercontent.com/assets/3941344/20683095/f012a29e-b5aa-11e6-9dac-53bb05623dac.png)

LDAP on mozillians.org and Google and GitHub login on AirMo have the same problem, both in normal and private browsing mode.

gene1wood:
> LDAP on mozillians.org ... have the same problem, both in normal and private browsing mode.

[...]

PR #81, opened 2016-11-30
In a condition, -1 does not equal "false" (only in a test)

gdestuynder:
Fixes https://github.com/mozilla-iam/auth0-deploy/issues/75
Will only trigger Duo when you are in the group that has Duo, instead of everyone in LDAP.

### DIFF
--- a/rules/duosecurity.js
+++ b/rules/duosecurity.js
@@ -3,7 +3,7 @@ function (user, context, callback) {
   // Please use http://github.com/mozilla-iam/auth0-rules instead
 
   // LDAP group okta_mfa requires MFA authentication everywhere.
-  if (user.groups && user.groups.indexOf("okta_mfa")) {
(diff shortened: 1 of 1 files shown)


rules/duosecurity.js
```diff
@@ -3,7 +3,7 @@ function (user, context, callback) {
   // Please use http://github.com/mozilla-iam/auth0-rules instead
 
   // LDAP group okta_mfa requires MFA authentication everywhere.
-  if (user.groups && user.groups.indexOf("okta_mfa")) {
+  if (user.groups && user.groups.indexOf("okta_mfa") !== -1) {
```
peterbe: It's quite common to use `> -1` instead.

gdestuynder:
true, > -1 or >= 0 seems a little more common. 
Also: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf

PR #83, opened 2016-11-30
Style: normalize use of Indexof

gdestuynder:
This should be all of them:

```
(master) ⚡ git grep indexOf
rules/application-access-override.js:  if (MOCO_MOFO_APPS.indexOf(context.clientName) >= 0) {
rules/application-access-override.js:        return user.groups.indexOf(group) >= 0;
rules/duosecurity.js:  if (user.groups && user.groups.indexOf("okta_mfa") >= 0) {
rules/mozdefqa-access.js:  if (APPS.indexOf(context.clientName) >= 0) {
rules/mozdefqa-access.js:        return user.groups.indexOf(group) >= 0;
(master) ⚡ 
```

### DIFF
--- a/rules/application-access-override.js
+++ b/rules/application-access-override.js
@@ -7,7 +7,7 @@ function (user, context, callback) {
   // LDAP groups allowed to access these applications
   var ALLOWED_GROUPS = ['team_moco', 'team_mofo'];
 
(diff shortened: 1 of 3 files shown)


gene1wood:
This cleans PR #81 which fixes issue #75

PR #84, opened 2016-11-30
IndexOf styling and fixes, mozdefqa ACL file

gdestuynder:
Please double check by testing before merging on testrp, and after merging as well (on the prod instance)


rules/mozdefqa-access.js
```diff
@@ -0,0 +1,27 @@
+function (user, context, callback) {
+  // /!\ DO NOT EDIT THIS FILE /!\
+  // Please use http://github.com/mozilla-iam/auth0-rules instead
+
+  // Applications that are restricted
+  var APPS = ['mozdefqa1.private.scl3.mozilla.com'];
+  // LDAP groups allowed to access these applications
+  var ALLOWED_GROUPS = ['vpn_mozdef'];
+
+  if (APPS.indexOf(context.clientName) >= 0) {
```
gene1wood: Since multiple clients can have the same name, we should in the long term use the client_id instead

PR #86, opened 2016-12-02
Add initial account linking support.

gdestuynder:
This is for DEV only and is here for review/comments/improvements.
i.e. this is not meant to be merged right now (even thus a copy of this exists in DEV).

### DIFF
--- a/rules/account-linking.js
+++ b/rules/account-linking.js
@@ -1,0 +1,95 @@
+/* vim: set ts=2 sw=2 et: */
+
+/* This rule merges the current account with an existing, verified, account
+* with the same email address.
+*
+* SECURITY CONSIDERATIONS
+* It is of the UPMOST IMPORTANCE that all emails used for merging
+* are marked as VERIFIED.
+*/
+
+/* TODO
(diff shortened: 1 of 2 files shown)


gdestuynder:
Current blocking issue found:

- when logging in as LDAP, it creates the LDAP account
- when logging in as Google afterwards, it links the account to the LDAP account (Primary: LDAP)
- login then fails because the Google account used to login was linked (and thus a temporary separated account was deleted)
- login again after the failure works (because the account is linked)

Potential fix:

- Call the API directly instead of through the auth0js lib and attempt to override the account currently logging in to the linked account instead of the "deleted" account, for example by forcing the user id. 
Just forcing the id while using auth0js lib did not work, thus I suspect there's more we can do.

If this cannot be fixed, this is a blocker and the automatic account linking would not be implemented in it's current form (ie we cannot do account linking without fixing this)

gdestuynder:
See also https://github.com/auth0/rules/issues/71

gdestuynder:
Closing for now as this implementation is not selected due to the UX problems (see previous comments)

PR #88, opened 2016-12-10
Change rules to clientid and update forbidden page

gene1wood:
# Forbidden page

This pairs with mozilla/sso-mozilla-com#4 and https://bugzilla.mozilla.org/show_bug.cgi?id=1322749
This will make it clear to the user what has happened

# Use clientID

This will make mapping these rules to clients more deterministic

### DIFF
--- a/rules/application-access-override.js
+++ b/rules/application-access-override.js
@@ -25,9 +25,13 @@ function (user, context, callback) {
     if (groupHasAccess) {
      return callback(null, user, context);
     } else {
-     // We use a redirect instead of UnauthorizedError() here as this ensure we can
-     // tell the user what happened. Otherwise, if the RP does not handle the error message
-     // there is no chance to tell the user why they're not getting logged in.
+     // Since this rule should only be used for RPs which can not do the
+     // authorization check themselves, and these types of RPs will likely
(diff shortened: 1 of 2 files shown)


rules/application-access-override.js
```diff
@@ -18,7 +26,7 @@ function (user, context, callback) {
      return callback(null, user, context);
     } else {
      context.redirect = {
```
gdestuynder: I wonder if we could now use UnauthorizedError return instead of a redirect, as it might take some time to have the page setup on sso.mozilla.com.
It used not to work but it's likely fixed.

Example:
return callback(new UnauthorizedError('You are not in the allowed group for this service'));

See also: https://auth0.com/docs/rules

gdestuynder: After testing: 
this is not the best way to do it. For RPs that do not implement handling of the error condition, this would not indicate to the user while they did not get logged in, neither  how to request access.
The redirect ensure that we keep control of the login flow and can display a page that will tell the user what happened, and potentially give a chance to request access.

gdestuynder:
@gene1wood r? (for subsequent changes)

PR #90, opened 2017-01-04
adding prod mozdef to access rule

jdow:


gdestuynder:
@jdow  there be conflicts. maybe this is already merged actually.

gdestuynder:
%jabba | kang: yeah, I think someone else added it already. can you delete it?

PR #91, opened 2017-01-11
Prevent invalid e-mails in the passwordless e-mail input from throwing errors in LDAP

### LINKED ISSUE #46: Putting an invalid email address in the passwordless field causes error message on login with LDAP
Go into a Lock 
Click Log In With Email
Put in some text (not a full, properly formatted e-mail address) 
Click Back
Click log in with LDAP
Put in LDAP creds
Try to click Log In. The CSS highlights the login button but no action takes place

Note that if you don't include a complete e-mail address in the LDAP e-mail field, clicking Login does give you a popup telling you to put in a valid e-mail address. So some logic is taking place. It just isn't completing the log in

tristanweir:
The Lock is designed with one Submit button, which behaves differently depending on if LDAP or Passwordless is activated. This causes unresolvable errors when a user put some text into the Passwordless input field, but then activates the LDAP login flow. 

This fix resolves the issue by clearing the Passwordless input field every time the user starts to go down the LDAP login flow. Fixes #46


gene1wood:
> The Lock is designed with one Submit button, which behaves differently depending on if LDAP or Passwordless is activated.

One point of clarification, the lock actually has two distinct submit buttons, one for LDAP and one for Passwordless. The right button is revealed to the user based on which type of auth they're attempting.

gene1wood:
Another approach to preventing the sending of multiple passwordless emails would be to, instead of modifying the DOM to disable the submit button (which may be slow), set a variable like `passwordlessSubmitClicked=true` and the in the `sendEmail` function check if that variable is set to prevent duplicate emails. I don't know if this speed benefit (of using a variable instead of modifying the DOM) is worth it.

PR #104, opened 2017-03-28
Changes the passwordless e-mail error message to prevent displaying u…

tristanweir:
Changes the passwordless e-mail error message to no longer display the error message. See https://bugzilla.mozilla.org/show_bug.cgi?id=1351213 for additional details

### DIFF
--- a/pages/login.html
+++ b/pages/login.html
@@ -297,7 +297,9 @@
         var icon;
         var backLink;
         if (err) {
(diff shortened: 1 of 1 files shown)


gene1wood:
Looks like this vuln came in from the [auth0 examples](https://github.com/auth0-samples/auth0-node-passwordless-sample/blob/4702141fa5ed2577effaf354d0e3125750b3d9a2/views/custom-link.jade#L26)

PR #108, opened 2017-03-31
Add initial version of CIS profile integration

gdestuynder:
This currently works if you manually provision the user profile into the
user's app_metadata structure in the auth0's management dashboard.
It will ignore users without app_metadata set.

Note: as per tech alignment meeting I added everyone as reviewer, though one single review suffices for merge. This also only merges to dev so far (though it shouldn't hurt prod if it ever gets carried over)

### DIFF
--- a/rules/profile_integration.js
+++ b/rules/profile_integration.js
@@ -1,0 +1,33 @@
+function (user, context, callback) {
+  // Currently, only run this rule if the user has app_metadata set
+  // And thus has the Mozilla user-profile configured
+  // Ideally the schema should be validated here as well for operational safety
+  if (typeof user.app_metadata !== 'undefined' && user.app_metadata) {
+    var extend = require('extend');
+    var user_profile = {};
+
+    // Preserve original profile data that we care for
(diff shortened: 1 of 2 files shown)


rules/profile_integration.js
```diff
@@ -0,0 +1,33 @@
+function (user, context, callback) {
+  // Currently, only run this rule if the user has app_metadata set
+  // And thus has the Mozilla user-profile configured
+  // Ideally the schema should be validated here as well for operational safety
+  if (typeof user.app_metadata !== 'undefined' && user.app_metadata) {
```
akatsoulas: nit: Probably this could be replaced with `if (user.app_metadata != null) {`.

gdestuynder:
This is the integration code API-side (which is going to be integrated in CIS) https://gist.github.com/gdestuynder/b08f032378c6196c9de5c579ddc5514b

gdestuynder:
Closing this, see https://github.com/mozilla/iam-project-backlog/issues/155#issuecomment-306961340

PR #109, opened 2017-04-04
Add rule support for HRData integration

gdestuynder:
This is a work-around.
Also, when the new user profile is implemented this data should be
sourced via CIS instead of LDAP, which will also be a little nicer/safer
since the data will not be included by default then wiped off the
profile, but only be included when necessary.

Note: the clientids are prod ids.

### DIFF
--- a/rules/hrdata.js
+++ b/rules/hrdata.js
@@ -14,7 +14,7 @@ function (user, context, callback) {
     context.samlConfiguration = context.samlConfiguration || {};
     //Remap SAML attributes as SAML cannot show Javascript objects
     for(var value in user._HRData){
-      var nname = "http://schemas.security.allizom.org/claims/HRData/"+value;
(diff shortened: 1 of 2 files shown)


rules/hrdata.js
```diff
@@ -0,0 +1,29 @@
...
+    'IU80mVpKPtIZyUZtya9ZnSTs6fKLt3JO', //biztera.com
+    'R4djNlyXSl3i8N2KXWkfylghDa9kFQ84' //mozilla.tap.thinksmart.com
+  ];
+
+  if (ALLOWED_CLIENTIDS.indexOf(context.clientID) >= 0) {
+    var extend = require('extend');
+    context.samlConfiguration = context.samlConfiguration || {};
+    //Remap SAML attributes as SAML cannot show Javascript objects
+    for(var value in user._HRData){
+      var nname = "http://schemas.security.allizom.org/claims/HRData/"+value;
```
gene1wood: Let's URL encode `value`

PR #112, opened 2017-04-17
adding thinksmart saml mappings

jdow:


gene1wood:
We really need to build a rule that consumes a data file with these kinds of things so that we remove our chances of a small code issue causing all users to be unable to auth to any RP. This (writing a custom rule for each exception) is a risky approach.

PR #115, opened 2017-06-01
adding rewrite for e-mail address for test google domain

jdow:


rules/test-mozilla-com-google.js
```diff
@@ -0,0 +1,15 @@
+function (user, context, callback) {
+  if (context.clientID !== 'q0tFB9QyFIKqPOOKvkFnHMj2VwrLjX46') return callback(null, user, context);
```
gdestuynder: for clarity, can you add the name of the clientid as a comment?

gdestuynder: also just add a general comment to explain what this rule is for

PR #119, opened 2017-06-09
fixing hrdata to include placeholder because can't be null/undefined.…

jdow:
… removing google from access override because shared accounts


rules/application-access-override.js
```diff
@@ -9,8 +9,6 @@ function (user, context, callback) {
     //'0123456789abcdefghijKLMNOPQRSTuv',  // auth : egencia.com
     //'123456789abcdefghijKLMNOPQRSTuvw',  // auth-dev : egencia.com
     'WXVdgVoCca11OtpGlK8Ir3pR9CBAlSA5', // auth : https://mozilla.slack.com
```
gdestuynder: Why are thees removed?

rules/irccloud-access.json
```diff
@@ -0,0 +1,4 @@
+{
+    "enabled": true,
+    "order": 7
```
gdestuynder: i think this has to be order 8 because test-mozilla-com-google.json is already 7 (ie it will break at deploy right now)

rules/application-access-override.js
jdow: Mainly because the shared accounts aren't in any groups that would allow them to log in. I went back and checked Okta and it did not have any restrictions, since Google has it's own access control and provisioning via GADS.

PR #123, opened 2017-06-14
Add cleanup rules for CIS profile (as much as can be cleaned up that is)

gdestuynder:
Fixes https://github.com/mozilla-iam/auth0-deploy/issues/122


gdestuynder:
I tested this in dev already

PR #124, opened 2017-06-21
Claim namespacing - automatically use the namespace for claims

gdestuynder:
In order to support OIDC Conformant mode in the future.

With this rule, non-OIDC conformant Clients (RPs) will get BOTH the old
and new claim which gives us some transition path forward.

**OIDC Conformant mode OFF before this rule:**
```
user = {
....
       "groups" = ["teddybear"],
.....
}
```
**OIDC Conformant mode OFF after this rule:**
```
user = {
....
       "groups" = ["teddybear"],
       "https://sso.mozilla.com/claim/groups" = ["teddybear"],
.....
}
```
**OIDC Conformant mode ON after this rule:**
```
user = {
....
       "https://sso.mozilla.com/claim/groups" = ["teddybear"],
.....
}
```

**TLDR**: as an RP, use the namespaced claims as soon as possible.

### DIFF
--- a/rules/claim_namespace.js
+++ b/rules/claim_namespace.js
@@ -1,0 +1,62 @@
+function (user, context, callback) {
+  // Auth0 in the "OIDC-Conformant" mode will only allow these claims. That mode
+  // enforces more than the OIDC spec in at least one regard:
+  // Auth0 forbids the use of optional claims UNLESS these are namespaced by a URL.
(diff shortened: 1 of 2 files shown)


gdestuynder:
NOTE: I have tested this in dev with both OIDC Comformant mode ON and OFF

PR #130, opened 2017-06-26
Merge of code already in prod

gdestuynder:
Fix was deployed during change window


gdestuynder:
FYI - @jdow @tyearns we'll need to talk to Biztera since their RP breaks with namespaced claims (we're good in prod as we tested and fixed during change window)

andrewkrug:
This still floats the button incorrectly at less than 748px.  @comzeradd any thoughts on how we can fix this?

gdestuynder:
i think the intention was that it only fails when you resize but not when you display

gdestuynder:
fixed

PR #135, opened 2017-07-04
Enable Google Tag Manager

comzeradd:


andrewkrug:
It adds google tag manager support for non-DNT clients.  Basically if I had to summarize the code it gives us:

* DNT Support 
* Click analytics
* Efficacy measurement on whether this feature is encouraging users to download nightly.

comzeradd:
What @andrewkrug said :)
In practice we want to track outbound links that people click (eg. which apps they mostly use).

PR #136, opened 2017-07-05
Enable GTM in prod

gdestuynder:
Dev was tested by @hmitsch 
Connectivity tests were run on OIDC/OIDC conformant/SAML


gdestuynder:
CHG0011907

PR #141, opened 2017-08-10
Update to newer lock and lock library, which fixes, the "remembered"

gdestuynder:
login for SAML initiated locks (for example this would trigger with
expensify and fail to login)

Note the new option is necessary as well because: https://github.com/auth0/lock/issues/1077

I tested all known combinations on testrp in dev

### DIFF
--- a/pages/login.html
+++ b/pages/login.html
@@ -363,8 +363,8 @@ C8Ar//Vn1zcq/Wwl6/8cfwO7+9oQfZdlAgAAAABJRU5ErkJggg==" alt="alert-info"><strong>W
  <![endif]-->
 
  <!-- This URL is to be manually updated -->
- <script src="https://cdn.auth0.com/js/lock/10.5/lock.min.js"></script>
- <script src="https://cdn.auth0.com/w2/auth0-7.4.min.js"></script>
+ <script src="https://cdn.auth0.com/js/lock/10.18/lock.min.js"></script>
+ <script src="https://cdn.auth0.com/w2/auth0-7.6.min.js"></script>
  <script>
     var APPS_WITH_TOS = [['phonebook-dev', 'https://www.mozilla.org/en-US/privacy/']];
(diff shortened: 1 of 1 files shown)


gene1wood:
This covers a lot of change, bringing the [lock from 10.5 to 10.18](https://github.com/auth0/lock/blob/master/CHANGELOG.md) and [auth0.js from 7.4 to 7.6](https://github.com/auth0/auth0.js/blob/v7/CHANGELOG.md). Have you read through the changelogs for these versions to know what will change with this upgrade?

gene1wood:
The [`emailInputPlaceholder`](https://github.com/mozilla-iam/auth0-deploy/blob/db417b0e268e97b2842e5658d01f659a0759cf84/pages/login.html#L632) is no longer respected. This is a regression

Viorela writes

Ldap login form is asking for user name and password, instead of email and password. By that I understand that, for example, for vioia@mozilla.com account the user name is vioia. 

I tried to login using just 'vioia', but the login failed.

emailinputplaceholder bug

gene1wood:
The [`error` argument](https://github.com/mozilla-iam/auth0-deploy/blob/db417b0e268e97b2842e5658d01f659a0759cf84/pages/login.html#L634-L638) is lost with this.

Viorela writes : "error message for invalid ldap credentials is changed (probably to fit the change above?)"

error message changed

gene1wood:
I'm not sure the root cause of this bug

Viorela writes "​When navigating to login page of reps-dev.allizom.org (which has ldap and passwordless login types available), the ldap login form is expanded by default and it cannot be collapsed. ​Trying to login with ldap when passwordless form is expanded results in browser console error. See: https://vimeo.com/229272072"

passwordless not working

gdestuynder:
These issues have been reported by Viorela by our asking following this PR. The merge is required for testing and this is exactly what the DEV environment is made for.
Thanks for reporting them on the PR though as these needs to be addressed.

PR #145, opened 2017-09-06
Enforce all logins through the GitHub connection/method to have 2FA e…

gdestuynder:
…nabled

@andrewkrug @comzeradd : i noticed that on sso.mozilla.com the /forbidden URL no longer exists (it used to exist somewhere at some point)
All rules that enforce access redirect to https://sso.mozilla.com/forbidden and just display a 404 instead.

Basically this means we probably want to also add that functionality to the SSO dashboard - woops.

The rule functionality otherwise works in the mean time, and is enabled on https://social-ldap-pwless.testrp.security.allizom.org/ for testing

### DIFF
--- a/rules/github_mfa.js
+++ b/rules/github_mfa.js
@@ -1,8 +1,13 @@
 function (user, context, callback) {
   if ((context.connection === 'github') && (!user.two_factor_authentication)) {
     console.log('GitHub user not allowed to log in because 2FA was disabled on the account: '+user.user_id);
+    var reason = 'You must setup a security device ("MFA", "2FA") for your GitHub account in order to access ' +
+      'this service. Please follow the ' +
+      '<a href="https://help.github.com/articles/securing-your-account-with-two-factor-authentication-2fa/"> ' +
+      'GitHub documentation</a> to setup the device and try logging in again.';
+
     context.redirect = {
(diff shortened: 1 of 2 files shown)


gdestuynder:
please review @andrewkrug

johngian:
Code looks OK the only comment i have is about the URL. Are we going to provide some more information about the reason we redirect to ``/forbidden`` ?

gdestuynder:
Rule updated and tested at: https://manage-dev.mozilla.auth0.com/#/rules/rul_YWLi9n2TXnMhUN6v

viorelaioia:
Small feedback on the forbidden page. I believe it is a bit confusing for a user to have "Return to dashboard" option when reaching the forbidden page in mozillians. I'd suggest to have instead a button that says something like "Try to login again" and redirects the user back to the login page for mozillians,

gdestuynder:
@viorelaioia  the rule cannot enforce this but it can pass a `client_id` or `redirect_uri` to the dashboard for example.
@andrewkrug @comzeradd  would that make sense for me to add? e.g. "sso.mozilla.com/forbidden?reason=...&from=https://www.mozillians.org/redirect_uri" for ex

andrewkrug:
@gdestuynder I think we should go ahead with this rule deployment as is.  I think we should file an enhancement to improve the args passed to the forbidden page.

PR #146, opened 2017-09-06
Enforce MFA for ALL ldap users

gdestuynder:
This REMOVES the `okta_mfa` group requirement to use Duo MFA on Auth0. Effectively this enforces Duo MFA when using Auth0 for ALL ldap user without any exception.

Should we audit which users are impacted and count them or are we just ok with that?
We can also instead send them to a page that say "create a GitHub account instead, your LDAP won't work anymore for this". Could be a "big" deal requiring comms.

(CC @jeffbryner @hmitsch because FYI/we forgot about that)


jeffbryner:
Doesn't sound like it will change anything for folks, but I agree a quick audit/count of who would be affected is a good idea.

hmitsch:
Will this break Viorela's test automation accounts?

gdestuynder:
@hmitsch no, @viorelaioia 's testing already uses MFA

gdestuynder:
can I reviews? @jdow

gdestuynder:
@hmitsch this blocks work until we have a decision

jeffbryner:
It's not clear to me what would be different for folks (except they need to setup MFA if they haven't) so it is that straightforward then; yes we should require MFA for all. 

Then the remaining issue is change management, for which it would help to know how many people are affected. Can we determine that?

gdestuynder:
last we checked, about 140 people, we have a list

hmitsch:
Inquiring with kang via direct message to get a more thorough understanding of the situation.

andrewkrug:
@gdestuynder @jeffbryner do we know what's blocking this at this point?  Is it filing the CAB or does it just need code review?

gdestuynder:
@andrewkrug code review would be a good start ! ;-) (really thats what blocking it, please do review :)

andrewkrug:
@gdestuynder this looks simple enough.  Let's go ahead and move forward with deploying this.

PR #147, opened 2017-09-12
Workaround for the updated_at attribute to display a timestamp instead

gdestuynder:
of a string, as per OpenID Connect spec.

Running/tested in dev for testrp. A test on sso dashboard would be good.


gdestuynder:
CHG0012092

PR #151, opened 2017-09-28
adding AWS consolidated billing mapping/access rules

jdow:
This probably won't be a permanent solution, but it's how Okta does SAML federation with AWS. The rules check whether the user is in the proper LDAP group, and then assigns the proper roles via the ARNs of the role and IDPs. This method uses one client per role. There is probably a better way, but for now, we just need to get off of Okta. These rules are live in production already, added via the auth0 dashboard. This commit is to ensure they stay there.

### DIFF
--- a/rules/AWS-consolidatedbilling-readonly.js
+++ b/rules/AWS-consolidatedbilling-readonly.js
@@ -1,0 +1,38 @@
+function (user, context, callback) {
+  var APPS = [
+    'koX1ze40wpoUovVV3RA7K79uTlxpbZFp' // AWS Consolidated Billing - Read Only
+  ];
+
+  var ALLOWED_GROUPS = ['aws_consolidatedbilling_read_only'];
+  
+  if (APPS.indexOf(context.clientID) >= 0) {
+    var groupHasAccess = ALLOWED_GROUPS.some(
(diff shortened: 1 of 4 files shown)


gdestuynder:
@jdow  @gene1wood aren't we going to go crazy if we do one rule per aws role per aws account ?!
Should we do this: https://github.com/mozilla-iam/iam-project-backlog/issues/210 right now instead, or are those 2 rules "needed now"?

jdow:
This is just for the two roles that were on Okta before. I know there is a lot of other AWS stuff in the works to make for a nicer environment, but perfection being the enemy of "good enough" is what led me to just implement exactly what Okta had going on, which I agree is terribly inefficient and non-pretty (because it's one client plus one rule for *every* role, even in a single account). I would suggest merging this, and noting the technical debt in an issue. In this case, we really aren't creating any new technical debt, it's just simply moving it from Okta to auth0.

gdestuynder:
wfm but would want @gene1wood 's thumb up just to be sure on the aws bits (i've not played with this side of auth0 as much)

gdestuynder:
@gene1wood im going to merge this though it would be good to ensure we have an issue to make it better

PR #152, opened 2017-10-02
Work around to reintegrate LDAP data correctly even when no

gdestuynder:
mozilliansorg_ event (or other publisher) exists. This is needed until
ther is a LDAP CIS publisher and supersdes CIS work-around

This fixes https://github.com/mozilla-iam/iam-project-backlog/issues/223

### DIFF
--- a/rules/cis.js
+++ b/rules/cis.js
@@ -59,12 +59,15 @@ function (user, context, callback) {
       console.log('CIS: HTTP Error while parsing user groups for user '+user.user_id+': '+err);
       return cb();
     });
+  } else {
+    return cb();
   }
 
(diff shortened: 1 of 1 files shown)


gdestuynder:
Code is in auth-dev already ("live")

andrewkrug:
This causes passwordless logins, google, and github to fail login with the error : 

```
Script execution did not complete within 20 seconds. Are you calling the callback function?
```

gdestuynder:
the issue is not ldap, but that the callback was not called if you had no `app_metadata` (log message is accurate)
fixed in dev and commit

andrewkrug:
+r

PR #154, opened 2017-10-05
Force users to login with their most secure login method.

gdestuynder:
Eg:
gdestuynder@mozilla.com is an LDAP account
gdestuynder@mozilla.com tries to login with passwordless, login fails
and user is told to login with LDAP

The order of preference is: passwordless < google < github < LDAP

### DIFF
--- a/rules/force-users-login-most-secure-method.js
+++ b/rules/force-users-login-most-secure-method.js
@@ -1,0 +1,84 @@
+function (user, context, callback) {
+  if (!user.email_verified) {
+    var reason = 'You primary email is not verified. Please contact EUS.';
+    context.redirect = {
+     url: "https://sso.mozilla.com/forbidden?reason="+encodeURIComponent(reason)
+    };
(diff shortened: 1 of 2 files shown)


gdestuynder:
this is the rule that was called "force v2" during development. it now has a proper name.

gdestuynder:
@viorelaioia Sorry, when i replied on https://github.com/mozilla-iam/iam-project-backlog/issues/223 i meant for this to be here. this is the actual thing to test (your tests are good/valid, i just put the comment at the wrong place)

andrewkrug:
We did test this at the Berlin work week.  It works as intended.  As a piece of feedback though we may want to do a better job with the messaging on the Forbidden page.  Currently, the language is not ideal and can be very confusing.  Otherwise the code reads pretty good top to bottom and was easy enough for me and @johngian to understand.

gdestuynder:
merging for now, happy to change language

PR #155, opened 2017-10-11
Update alerts colors

comzeradd:
This is based on the new Mozilla brand guidelines

### DIFF
--- a/pages/login.html
+++ b/pages/login.html
@@ -15,10 +15,26 @@
         text-align: center;
         display: none;
         font-family: "Avenir Next", Avenir, -apple-system, BlinkMacSystemFont, Roboto, Helvetica, sans-serif;
-      }
-      .alert-info {
-        background-color: #538195;
         color: white;
+      }
+      .alert-success {
+          background-color: #14c887;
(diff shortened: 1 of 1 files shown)


gdestuynder:
lets hold onto that until this Friday in case things change (see https://www.pivotaltracker.com/story/show/152020211)
Note: we could probably also drop the alert from the lock at this point, I think, since dashboard also has alerts

gdestuynder:
@comzeradd so this actually changed ;-)
I wonder if we should just drop the alert, or wait for the new lock?

comzeradd:
I'd drop that alert from the lock. We are going to drop this alert anyway when F57 is released.
When we have the new lock we can rethink if we want to show alerts there or not.

gdestuynder:
@jeffbryner i can drop the beta alert during the prod cab tomorrow, if we want. need your ack though!

jeffbryner:
I think it's fine to remove it since we have the SSO dashboard alerting.

gdestuynder:
ok, will do!
closing pr

PR #158, opened 2017-10-18
Fix logic to always select the best account, not the "next-best".

gdestuynder:
i.e. if you had passwordless, github and LDAP, and you logged in as
passwordless, ie would select github first and tell you to login as
github.
Then you login as github, and it would select LDAP and tell you to login
as LDAP. Then finally, you login as LDAP and it works. Terrible UX :)

This code will now directly select LDAP in this case and tell you to
login as LDAP, regardless of you trying to login as passwordless or
GitHub first.

Additionally, this now sends the `redirect_uri` with the error message,
as a parameter to the SSO Dashboard, which may make use of this in the
future to send the user back.

### DIFF
--- a/rules/force-users-login-most-secure-method.js
+++ b/rules/force-users-login-most-secure-method.js
@@ -80,7 +80,6 @@ function (user, context, callback) {
         if (err) {
           callback(err, user, context);
         } else { // No error, but loop ended
-          console.log(context);
           console.log('User profile that may log in is: '+selected_user.user_id+' initial login attempt was with: '+user.user_id);
             if (user.user_id !== selected_user.user_id) {
               var reason = 'Sorry - you may not login with that user account. Please always' +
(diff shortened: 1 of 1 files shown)


gdestuynder:
cc @comzeradd for info (on redirect_uri being passed to the dashboard when we redirect the user)

ps: code is live in dev.

rules/force-users-login-most-secure-method.js
```diff
@@ -62,31 +62,34 @@ function (user, context, callback) {
...
           // XXX This currently assumes single identity/no linked account /!\
           var provider = targetUser.identities[0].provider;
-          console.log(targetUser.user_id+' match order '+matchOrder[provider]);
-          if (matchOrder[provider] < matchOrder[myprovider]) {
-            console.log('Found better account target');
+          var previous_provider = selected_user.identities[0].provider;
+          if (matchOrder[provider] < matchOrder[previous_provider]) {
             selected_user = targetUser;
           }
+          console.log(targetUser.user_id+'is of match order '+matchOrder[provider]+'. Selecting: '+selected_user.user_id);
```
johngian: Please add extra space before ``is`` to make logs more readable.

PR #159, opened 2017-10-18
Always return if there are no LDAP groups to process, otherwise our

gdestuynder:
webtask will die and cause a login failure

### DIFF
--- a/rules/cis.js
+++ b/rules/cis.js
@@ -30,9 +30,14 @@ function (user, context, callback) {
     function(err, response, body) {
       if (!err) {
         var udata = JSON.parse(body);
+        // If user has no LDAP groups, return immediately
+        if (udata.groups === undefined) {
+          return cb();
+        }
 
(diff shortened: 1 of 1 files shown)


rules/cis.js
```diff
@@ -30,9 +30,14 @@ function (user, context, callback) {
     function(err, response, body) {
       if (!err) {
         var udata = JSON.parse(body);
+        // If user has no LDAP groups, return immediately
+        if (udata.groups === undefined) {
+          return cb();
```
akatsoulas: What's this `cb()` call?

gdestuynder: its defined here: https://github.com/gdestuynder/auth0-deploy/blob/0cfea013ee73cfeeedb0d7df7aa4e34a1296630d/rules/cis.js#L72

`cb` stands for `callback` (the later is already used in the scope of the webtask function, ie "our" global scope)
It performs final checks and calls the real `callback` function

PR #170, opened 2018-02-09
add access rule for IAM AWS Console access

andrewkrug:


gdestuynder:
@andrewkrug is this rule permanent and does it supersedes https://github.com/mozilla-iam/auth0-deploy/blob/master/rules/SAML-temporary-AWS-consolidatedbilling-admin.js ?
Should it be driven by apps.yml (or similar) instead of hardcoded?

gene1wood:
Do you have any interest in updating [the existing rule](https://github.com/gene1wood/auth0-federated-aws/blob/master/auth0-rules/auth0-rule-aws-federated-group-role-mapping.js) instead of creating a second one?

https://github.com/gene1wood/auth0-federated-aws/

This would keep the complexity down and allow for a single global Auth0 client

gene1wood:
I'd also like to move the [git backed rule](https://github.com/gene1wood/auth0-federated-aws/blob/master/auth0-rules/auth0-rule-aws-federated-group-role-mapping.js) into a place in auth0-deploy now that the deploy plugin is working again.

andrewkrug:
@gene1wood https://github.com/gene1wood/auth0-federated-aws/ --> is not in auth0 production.  Is there a reason why we couldn't replace all of the AWS account access rules with this rule?

andrewkrug:
@gdestuynder -- re: @andrewkrug is this rule permanent and does it supersedes https://github.com/mozilla-iam/auth0-deploy/blob/master/rules/SAML-temporary-AWS-consolidatedbilling-admin.js ?
Should it be driven by apps.yml (or similar) instead of hardcoded?

> The answer is it's temporary and can supersede the others if we adopt the rule that gene proposed above.

gene1wood:
The benefit of putting the account to AWS saml-provider mapping and the IAM role to LDAP group mapping in a json data structure is that it sets us up for federating this down the road.

See [this diagram](https://github.com/gene1wood/auth0-federated-aws/#diagram) with an example of making the group to role mapping self service and note centralized in our rule.

In this PR the data of these two mappings is embedded in the code.

The additional logic that's in your PR is the ability to bar users who are not member of a group in `ALLOWED_GROUPS` from attempting to do federated login. Is this needed? Wouldn't the group to role mapping take care of this such that a user which was not part of any LDAP group that mapped to an IAM role would have no possible IAM roles to assume and so would be barred from accessing?

The other goal of that rule I have in dev (which we can easily deploy to prod if we wish) is to standardize on 2 global Auth0 client_ids for all AWS accounts federated access (stage and prod) instead of one client per AWS account.

Could we just add in the mapping that you want and deploy [this rule](https://github.com/gene1wood/auth0-federated-aws/blob/master/auth0-rules/auth0-rule-aws-federated-group-role-mapping.js) in prod?

The mapping change would look like this :

```
--- <unnamed>
+++ <unnamed>
@@ -10,7 +10,8 @@
   // Test for content in the storage cache and re-fetch if missing
   var DATA = {
     "accounts": {
-      "656532927350" : "arn:aws:iam::656532927350:saml-provider/Mozilla-Auth0-Dev"
+      "656532927350" : "arn:aws:iam::656532927350:saml-provider/Mozilla-Auth0-Dev",
+      "320464205386" : "arn:aws:iam::371522382791:saml-provider/Mozilla-Auth0-Prod"
     },
     "account_role_group_mapping" : {
       "656532927350" : {
@@ -20,7 +21,11 @@
         "arn:aws:iam::656532927350:role/ReadOnlyFederated" : [
           "team_opsec"
         ]
-
+      },
+      "320464205386" : {
+        "arn:aws:iam::320464205386:role/FederatedAWSAccountRead" : [
+          "mozilliansorg_mozilla-iam-aws-access"
+        ]
       }
     }
   };
```

gene1wood:
I may have misunderstood your question

> s there a reason why we couldn't replace all of the AWS account access rules with this rule?

As I wasn't sure which rule was `this rule`

PR #171, opened 2018-02-15
fxa script support

gdestuynder:
image 1
Support for this


gdestuynder:
Note, 2FA support is required before we enable in prod, though since this in /manual no problem merging

PR #172, opened 2018-02-15
Add LDAP-check, make loading state first state

hidde:
Add LDAP check:

* [Try Person API if email is not definitely LDAP](https://github.com/mozilla-iam/auth0-custom-lock/commit/7e15a566d7eda101d656b8b96f77ecedd30abccb)
* [Replace LDAP check by isDefinitelyLDAP check](https://github.com/mozilla-iam/auth0-custom-lock/commit/ea15a525756a8f9d9f426ef5e2fee7123c3a4a7c) (if it matches our regex, we can be certain it is LDAP and don't need to call Person API)

Make loading state first state:

* [Make loading state the very first](https://github.com/mozilla-iam/auth0-custom-lock/commit/0e869cac2d9b83abf1abec8349cad852706ddde6) (before, user would see: form, then loading because checking RP connections, then form again; now: user just sees loading state, then form again)

### DIFF
--- a/pages/login.html
+++ b/pages/login.html
@@ -48,7 +48,8 @@
       input[type="email"]:focus,
       input[type="password"]:hover,
       input[type="password"]:focus {
-        border: 1px solid #229DC4; }
+        border: 1px solid #229DC4;
+        outline: none; }
       input[type="text"]:focus,
       input[type="email"]:focus,
       input[type="password"]:focus {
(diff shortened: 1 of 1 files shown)


gdestuynder:
https://github.com/mozilla-iam/auth0-custom-lock/commit/7e15a566d7eda101d656b8b96f77ecedd30abccb#diff-7cc1d4e217e5bc83372d495f039a0562R8
this probably needs to hit prod, otherwise we need a staging branch

gdestuynder:
https://github.com/mozilla-iam/auth0-custom-lock/commit/ea15a525756a8f9d9f426ef5e2fee7123c3a4a7c note we also use mozillafoundation.org though id rather let person-api decide so that this code is only in one place

PR #175, opened 2018-02-15
Add default-off rule that denies login for maintenance

### LINKED ISSUE #174: Add default deny rule to auth0 rules
Currently if all auth0 rules were deleted in order to resolve an ordering issue (like what we're doing currently in dev) the absence of any rules would cause sites that use things like [`AccessRules`](https://github.com/mozilla-iam/auth0-deploy/blob/master/rules/AccessRules.js) to allow all authenticated users into the site regardless of group membership. Currently and in the future there could be other rules which effectively deny users from accessing something.

In order to mitigate against this risk, let's make auth0 rules work more like firewally rules by adding a "default deny" rule to run at the very end of the rule order. This would mean that if all other rules were deleted, other than this "default deny" rule, users would be barred from logging in to any site.

The assumption here is that it's better to bar everyone from accessing any site, than to allow an unauthorized user from accessing a site they shouldn't be allowed to.

@gdestuynder suggested this could be accomplished by creating a "default deny" rule and putting it in order position `999` or some such large number.

gdestuynder:
Add default-off rule that denies login for maintenance. Order is 999.
Only use this when required as it forbids ALL logins.

NOTE: This introduce a new message 'maintenancemode' which needs to also
be present on the sso dashboard (it works without, but looks prettier
with)

Fixes #174

### DIFF
--- a/rules/default-deny-for-maintenance.js
+++ b/rules/default-deny-for-maintenance.js
@@ -1,0 +1,6 @@
+function (user, context, callback) {
+    // Denies all users from logging in
+    // Only use for maintenance purposes
+    
+    return callback(null, user, global.postError('maintenancemode', context));
+}
--- a/rules/default-deny-for-maintenance.json
+++ b/rules/default-deny-for-maintenance.json
@@ -1,0 +1,4 @@
+{
+    "enabled": false,
+    "order": 999
+}


gdestuynder:
correct, `enabled = true` is required to enable (or flip it in the dashboard)
otherwise, it will always deny so that'd be a problem

gene1wood:
What do you think about moving more to how firewall rules work where a positive indicator is needed to pass through the rules. So for example, right now where we check to see if the client_id is one that we know the auth0 rules need to do authorization checks for (because the RP doesn't have the capability), what if in that client_id check, when we find that the client_id does *not* need our help, we set a "yes, this user should be allowed to login flag which gets us out of the rule processing flow. This way, just like with firewall rules, we can leave the 999 rule enabled and then when we make a mistake in the rules (related to their order, a bug in a rule, etc) the default behavior is not to allow the user to login, but instead to bar the user from logging in.

gdestuynder:
while this sounds good in theory, we're limited to how auth0 is implemented and I would be careful when implementing this as there is no state guaranteed for volatile data between rules.
ie, auth0 is not currently made for this - and the ways I can think of to implement might break unknowingly  with updates - which would fail safe (forbid login) but also be terrible from an availability stand point

f.e.:
- we can try to create `context.login_allowed` and hope this keeps working
- Another possibility is to create `app_metadata.login_allowed` and delete it at the 999 rule, and similarly hope this keep working (there chances this shows in the user profile but maybe that's fine)

The advantage of the current rule is that its not state dependent basically.

I propose we just ask Auth0 what they think of this and what their recommendation is

gdestuynder:
Bundling this in #176

PR #191, opened 2018-04-18
Hook for post-user which sends the user_id of the created users to an

gdestuynder:
API
Supports cloudwatch events

This is what's running at https://manage-dev.mozilla.auth0.com/#/hooks


gene1wood:
Out of curiosity is there a link to the code that generates the associated AWS IAM user/role that the AWS access key is associated with?

gdestuynder:
no, its not yet generated

gdestuynder:
we're not currently using this because the data we want is not present in auth0 hooks at this time

PR #192, opened 2018-05-02
Add pto app to hrdata rule

jdow:
this will allow the pto app to get manager info and no longer need an LDAP connection


gdestuynder:
@jdow could this app use hris_ group assertions instead? we want to retire 'HRDATA'

hmitsch:
In the Workday export we seem to have `Worker_s_Manager_s_Email_Address`. Is this something we want to add to the OIDC assertions? @gdestuynder

gdestuynder:
yup

hmitsch:
Should I add this to the *draft* backlog for our next sprint? @gdestuynder

gdestuynder:
backlog yes, but i dont think itll be next sprint

PR #194, opened 2018-05-17
Support for expiration of access

gdestuynder:
Support for expiration of access as per https://github.com/mozilla-iam/cis/blob/master/docs/rfcs/AutomaticAccessExpiration.md Note that the users MUST have an authoritativeGroups entry for the clientid (uuid) for the access to be updated (authoritativeGroups.lastUsed) and granted/denied.

This means code in the Auth0 publisher needs to be added to detect when
the value is setup and add the attribute to the corresponding users ; or
; default setup `lastUsed` on the first login attempt (which has the
vulnerability of being able to "never login" then 10000 days later still
be able to login for the first time. This is less of an issue for RPs
that would always have had this setup of course.

Note also that this is fully independent from access groups and any
other access, and overrides every other access when setup.

See also: https://github.com/mozilla-iam/sso-dashboard-configuration/pull/148


gdestuynder:
This is live in dev.

gdestuynder:
tested all scenarios:
- rp has no settings
- rp has authorized user
- rp has authorized groups
- rp has expiration setting
- mix of all rp settings
- user is expired
- user is not expired
- user has no setting

Seems good to go

PR #197, opened 2018-06-04
force using what mozilliansorg considers to be the primary account this is a work-around which should be removed in the future

gdestuynder:
This is the enforcing side of mozilliansorg forcing a primary account/login method. More words:

- Mozillians.org considers accounts by both `user_id` and `primaryEmail` and will only push group updates for one of these ("your primary account")
- Mozillians.org will not consider an account with the same `primaryEmail` but different `user_id` to be different accounts
- Mozillians.org (staging) lets you selected a "primary account" which call the auth0 api in order to inform auth0 of which account is "primary" and should be allowed in. This is because otherwise the group data may be inconsistent (due to the points above)

This rule enforces Mozillians.org's view of the primary account so that there is no inconsistency.

IMPORTANT NOTES:

This allows auth0 to bypass CIS check as long as this is in place at it can override other `user.app_metadata` values and has to be changed in the future. This is a security thing (@andrewkrug @gene1wood )

The user should be directed to the login again with the correct user (eg `/login....&login_method=x&identifier=z&reason=notprimary` instead of sending the user to an error page, or to mozillians.org for the same reason. That said for now, its the error page. (this is a UX thing @hidde @m-branson )

The above code assumes that mozillians.org WILL correctly set `user.app_metadata.mozilliansorg_primary` and correctly UPDATE it (ie false vs true). If this is not the case, the code would have to be modified (consider github + no data to be "true"?) @akatsoulas @johngian

### DIFF
--- a/rules/force-users-login-most-secure-method.js
+++ b/rules/force-users-login-most-secure-method.js
@@ -73,7 +73,9 @@ var WHITELIST = ['HvN5D3R64YNNhvcHKuMKny1O0KJZOOwH', // mozillians.org account v
     if (data.length > 0) {
       // Initialize selected_user as current user
       var selected_user = user;
+      selected_user.app_metadata = selected_user.app_metadata || {};
       async.each(data, function(targetUser, cb) {
+        targetUser.app_metadata = targetUser.app_metadata || {};
         // Only list "not us" ;-)
         if (targetUser.user_id !== user.user_id) {
           // XXX This currently assumes single identity/no linked account /!\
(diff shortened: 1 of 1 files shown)


gdestuynder:
Note, this is in auth0-dev

gdestuynder:
corresponding error msg https://github.com/mozilla-iam/sso-dashboard/pull/309

gdestuynder:
@akatsoulas note we plan to can this, check with @hmitsch before reviewing

gdestuynder:
closing for now

PR #206, opened 2018-07-12
last-resort maintenance page for login

gdestuynder:
screenshot_20180712_100632

### DIFF
--- a/manual/maintenance.html
+++ b/manual/maintenance.html
@@ -1,0 +1,19 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <title>Mozilla Login - Maintenance</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+  </head>
+  <body>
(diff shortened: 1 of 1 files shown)


manual/maintenance.html
```diff
@@ -0,0 +1,28 @@
...
+  <body>
+    <h1>Maintenance mode</h1>
+    <p>
+    We are currently experiencing an issue &amp; working hard to resolve it. Please try to login again in a few minutes.
+    </p>
+    <br/>
+    <p>
+    For emergencies, you can request help via <a href="mailto:moc@mozilla.com">moc@mozilla.com</a>, on IRC channel <a
+    href="irc://irc.mozilla.org/moc">#moc</a>
+    or Slack channel <a href="https://mozilla.slack.com/">#servicedesk</a>
```
hidde: nice! quickly tried it we could do slack links but seems that doesn’t reliably so this is great!

PR #208, opened 2018-07-26
KR20 Signing support for apps.yml

gdestuynder:
This is the auth0 piece that verifies a JWT signed version of apps.yml
It assumes that apps.yml itself is a JWT and thus contains all data (ie the b64 encoded JWT) in it's current form, though we could decide to do it differently if we wanted.

NOTE: This only works on newer auth0 with nodejs8 (i took the opportunity to refactor stuff and use `await` and what not. Yay nodejs8!)

NOTE2: since there's no well-known endpoint and we don't yet have a testsuite for rules, this is obviously not tested against a simili or real env, just local 

NOTE3: DO NOT DEPLOY until the well-known endpoint is available and the apps.yml signed, even in dev since that'll fail verification and fetch :)


gdestuynder:
Also removed the now unused rule, forgot it was there ;-)

rules/AccessRules.js
```diff
@@ -1,7 +1,58 @@
...
+    // terminated. Generally this means we hit the same task within 60s.
+    if (global.access_rules) {
+      return global.access_rules;
+    }
+
+    try {
+      var options = { method: 'GET', url: access_file.endpoint };
+      request(options, function (error, response, body) {
+        if (error) throw new Error(error);
+        var decoded = jwt.verify(body, configuration.iam_jwt_rsa_pkey, function(err, decoded) {
```
gdestuynder: this part should get jwt rsa pkey from well-known when available

gdestuynder:
now uses JWS

PR #212, opened 2018-08-06
Add common voice dev to ratcheting bypass

gene1wood:
This is to facilitate testing Common Voice using passwordless and LDAP without ratcheting.

I'm, at the moment, confirming with @hidde that the Auth0 application (previously called Client or RP) is setup correctly.


gene1wood:
I'll close this as my commit wasn't signed for some reason and we can try this again.

PR #222, opened 2018-10-03
No ratcheting code

gdestuynder:
This contains the DEV version of:

- [x] Remove login ratcheting entierely
- [x] Enforces LDAP logins for all LDAP accounts (ie Google=>LDAP logins do not work anymore)
- [x] Supports AAI (See https://github.com/mozilla-iam/cis/blob/profilev2/docs/AuthenticatorAssuranceIndicator.md)
- [x] Default AAI level requirements to MEDIUM for apps which do not have a setting in the access file
- [x] Remove the classic MFA enforcement as AAI supersedes that concept

What isn't there:
- [ ] Redirect error msgs to NLX (though this work, but NLX gets overwritten, so we need support for this in NLX code to put that in our code)


gdestuynder:
Note: you can play with this on testrp, and if you check your claims, AAI and AAI_LEVEL should now be present.

- AAI is your set of indicators for that user
- AAI_LEVEL is the level that was required for you to login to that RP

PR #226, opened 2018-10-11
Change the behavior to decide whether or not to set the AAI 2FA attribute as present

gdestuynder:
Change the behavior to decide whether or not to set the AAI 2FA attribute as present.

Previously `2FA` was set If the `user.multifactor` attribute is set, which meant the user *had* authenticate**d** with duo.

This change adds a case so that if `context.multifactor` is set, it means the user will be required to duo-authenticate next, and as a result the `2FA` AAI attribute is set

see the duosecurity rule if curious

cc @viorelaioia

### DIFF
--- a/rules/aai.js
+++ b/rules/aai.js
@@ -7,7 +7,7 @@ function (user, context, callback) {
     Array.prototype.push.apply(user.aai, ["2FA"]);
   } else if ((context.connection === 'firefoxaccounts') && (user.fxa_twoFactorAuthentication === true)) {
     Array.prototype.push.apply(user.aai, ["2FA"]);
-  } else if ((context.connectionStrategy === 'ad') && (user.multifactor[0] === "duo")) {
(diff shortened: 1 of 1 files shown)


gene1wood:
To be explicit, it sounds like this change results in
* When a user first attempts to login to Auth0 with LDAP (a new LDAP user)
* They show up with `context.multifactor` set to True (but no `user.multifactor` value yet) which means that at the end of the rule flow, they'll be sent on to Duo
* Before that point, the AAI 2FA will be set on their profile
* They'll then be sent off to Duo
* At this point if they correctly enter their Duo code, they come back and the speculative AAI 2FA attribute that was already set will have been correct
* If however they don't enter their Duo code correctly (or they're an attacker and don't have the 2FA device), though the AAI 2FA attribute will be set, implying that they've provided 2FA, they won't be able to finish authenticating and so no RP should ever see that misleading AAI 2FA attribute

PR #229, opened 2018-11-05
update rules to match production

andrewkrug:


rules/gcp-gsuite-SAML-claims.js
```diff
@@ -0,0 +1,24 @@
+function (user, context, callback) {
```
gdestuynder: @andrewkrug is this duplicating https://github.com/mozilla-iam/auth0-deploy/blob/master/rules/SAML-test-mozilla-com-google.js ?

rules/SAML-AWS-mozillaiam-account-readonly.js
```diff
@@ -0,0 +1,52 @@
+function (user, context, callback) {
+  var CLIENT_IDS = [
```
gdestuynder: faster would be   `if (context.clientID !== '2lTZtlpqx4bslng167l1BBqTMusCAJkZ') return callback(null, user, context); // AWS Federated Access for AWS IAM`

rules/gcp-gsuite-SAML-claims.json
```diff
@@ -0,0 +1,4 @@
+{
+    "enabled": true,
+    "order": 20
```
gdestuynder: this is already taken, please use `30` f.e.

rules/gcp-gsuite-SAML-claims.js
andrewkrug: It does the same replace.  But for a different namespace / client_id.

rules/gcp-gsuite-SAML-claims.json
andrewkrug: fixed

PR #236, opened 2018-12-10
Add automatic user profile linking support

gdestuynder:
This will:

- link users with a matching `primaryEmail` into a single user profile
- use the user_id of the previously existing profile, in a cascading
order. Eg for the SAME `primaryEmail`:
  - you logged in with GitHub first pre-linking code
  - you logged in with LDAP next
  - Post-linking code, your user_id will be your GitHub user_id and both
profiles will be linked. Data of both profiles will be merged.
- Profile data is merged (including metadata)
- Original profile data will be retained in the `identities`
  sub-structure and restored if accounts are de-linked, except for
metadata
- When linked, profiles will provide the *same* group structure (i.e.
same access)
- If an AAL LOW and MEDIUM (for example) are linked, AAL properties will
still be respected (ie if you login with passwordless to an RP requiring
AAL MEDIUM, while your passwordless account is linked to a 2FA'd GitHub
account your access WILL BE DENIED, and you will have to login using
GitHub)

**WARNINGS HERE** (ie things I know might be issues)

- Yes, this means you can end up with a passwordless `user_id` for an FxA account for example. I may have to add ratcheting logic in there if we foresee an issue with the `user_id` matching

- Account verification in Mozillians.org might be broken by this. Please test and let me know, so we can check if there is a work-around on the auth0 side. If note, we'd have to finally fix the verification in Mozillians.org/DinoPark to *not* use Auth0 directly.

### DIFF
--- a/rules/link-users-by-email-with-metadata.js
+++ b/rules/link-users-by-email-with-metadata.js
@@ -18,6 +18,17 @@ function (user, context, callback) {
     return callback(null, user, context);
   }
 
+  // Lower is better - used by linking ratcheting logic
+  const matchOrder = {'Mozilla-LDAP': 0,
+                      'Mozilla-LDAP-Dev': 0,
+                      'firefoxaccounts': 1,
+                      'github': 2,
+                      'google-oauth2': 3,
+                      'email': 4
+                     };
+  // How old an account should be to be considered "new" to the system for linking, in seconds
+  const user_ratchet_link_delay_sec = 300;
+
   const userApiUrl = auth0.baseUrl + '/users';
   const userSearchApiUrl = auth0.baseUrl + '/users-by-email';
 
@@ -40,20 +51,36 @@ function (user, context, callback) {
       return u.email_verified && (u.user_id !== user.user_id);
     });
 
-//    if (data.length > 1) {
-      // We already have  more than one match, so we need to decide which account is going to be main
-      // This code is only triggered by accounts that were present before automatic linking was enabled
-      
-//    }
     if (data.length === 0) {
       // Already linked or no matching email found, continue with login
       return callback(null, user, context);
     }
 
-    const originalUser = data[0];
+    // Default to ourselves as main profile
+    var originalUser = user;
+
+    // We have matches in Auth0, so we need to decide which account is going to be main
+    // Please see https://github.com/mozilla-iam/mozilla-iam/blob/master/docs/deratcheting-user-flows.md#user-logs-in-with-the-mozilla-iam-system-for-the-first-time 
+    // for detailed explanation of what happens here.
+    // NOTE: No date time check is made here as the current/new user is not yet present in Auth0 database (anymore?)
+    // thus we do not need to verify that the current profile (`user`) has a "new" `user.created_at` value. Yay!
+    for (var i = 0, len = data.length; i < len; i++) {
+      var targetUser = data[i];
+
+      var otherConnection = targetUser.identities[0].connection;
+      var curConnection = originalUser.identities[0].connection;
+      console.log(otherConnection, curConnection);
+
+      if (matchOrder[otherConnection] < matchOrder[curConnection]) {
+        console.log("Found user_id that should be main profile used for linking, according to ratcheting logic: " + targetUser.user_id);
+        originalUser = targetUser;
+      }
+    }
+
+    // Current user providers
     const provider = user.identities[0].provider;
     const providerUserId = user.identities[0].user_id;
-    
+
     console.log("Performing automatic profile linking: main profile: "+originalUser.user_id+" is now also main for: "+user.user_id);
 
     user.app_metadata = user.app_metadata || {};
--- a/rules/link-users-by-email-with-metadata.js
+++ b/rules/link-users-by-email-with-metadata.js
@@ -81,7 +81,7 @@ function (user, context, callback) {
     const providerUserId = user.identities[0].user_id;
 
     if (originalUser.user_id === user.user_id) {
-      // The profile we're trying to link is the same as the one we're login as
+      // The profile we're trying to link is the same as the one we're logged in as
       // This happens if the linking ratcheting logic selects the same profile and other profiles are not yet linked
       // These will be linked the first time the user login with them instead
       console.log("No automatic profile linking performed due to main profile matching current profile for: " + user.user_id);
--- a/rules/link-users-by-email-with-metadata.js
+++ b/rules/link-users-by-email-with-metadata.js
@@ -69,7 +69,6 @@ function (user, context, callback) {
 
       var otherConnection = targetUser.identities[0].connection;
       var curConnection = originalUser.identities[0].connection;
-      console.log(otherConnection, curConnection);
 
       if (matchOrder[otherConnection] < matchOrder[curConnection]) {
         console.log("Found user_id that should be main profile used for linking, according to ratcheting logic: " + targetUser.user_id);
@@ -81,7 +80,15 @@ function (user, context, callback) {
     const provider = user.identities[0].provider;
     const providerUserId = user.identities[0].user_id;
 
-    console.log("Performing automatic profile linking: main profile: "+originalUser.user_id+" is now also main for: "+user.user_id);
+    if (originalUser.user_id === user.user_id) {
+      // The profile we're trying to link is the same as the one we're login as
+      // This happens if the linking ratcheting logic selects the same profile and other profiles are not yet linked
+      // These will be linked the first time the user login with them instead
+      console.log("No automatic profile linking performed due to main profile matching current profile for: " + user.user_id);
+      return callback(null, user, context);
+    }
+
+    console.log("Performing automatic profile linking: main profile: "+originalUser.user_id+" is now also main for: " + user.user_id);
 
     user.app_metadata = user.app_metadata || {};
     user.user_metadata = user.user_metadata || {};
--- a/rules/link-users-by-email-with-metadata.js
+++ b/rules/link-users-by-email-with-metadata.js
@@ -1,0 +1,83 @@
+/**
+ * @title Link Accounts with Same Email Address while Merging Metadata
+ * @overview Link any accounts that have the same email address while merging metadata.
+ * @gallery true
+ * @category access control
+ *
+ * This rule will link any accounts that have the same email address while merging metadata.
+ * Source/Original: https://github.com/auth0/rules/blob/master/src/rules/link-users-by-email-with-metadata.js
+ *
+ */
+
+function (user, context, callback) {
+  const request = require('request');
+
+  // Check if email is verified, we shouldn't automatically
+  // merge accounts if this is not the case.
+  if (!user.email || !user.email_verified) {
+    return callback(null, user, context);
+  }
+
+  const userApiUrl = auth0.baseUrl + '/users';
+  const userSearchApiUrl = auth0.baseUrl + '/users-by-email';
+
+  request({
+   url: userSearchApiUrl,
+   headers: {
+     Authorization: 'Bearer ' + auth0.accessToken
+   },
+   qs: {
+     email: user.email
+   }
+  },
+  function (err, response, body) {
+    if (err) return callback(err);
+    if (response.statusCode !== 200) return callback(new Error(body));
+
+    var data = JSON.parse(body);
+    // Ignore non-verified users and current user, if present
+    data = data.filter(function (u) {
+      return u.email_verified && (u.user_id !== user.user_id);
+    });
+
+//    if (data.length > 1) {
+      // We already have  more than one match, so we need to decide which account is going to be main
+      // This code is only triggered by accounts that were present before automatic linking was enabled
+      
+//    }
+    if (data.length === 0) {
+      // Already linked or no matching email found, continue with login
+      return callback(null, user, context);
+    }
+
+    const originalUser = data[0];
+    const provider = user.identities[0].provider;
+    const providerUserId = user.identities[0].user_id;
+    
+    console.log("Performing automatic profile linking: main profile: "+originalUser.user_id+" is now also main for: "+user.user_id);
+
+    user.app_metadata = user.app_metadata || {};
+    user.user_metadata = user.user_metadata || {};
+    auth0.users.updateAppMetadata(originalUser.user_id, user.app_metadata)
+    .then(auth0.users.updateUserMetadata(originalUser.user_id, user.user_metadata))
+    .then(function() {
+      request.post({
+        url: userApiUrl + '/' + originalUser.user_id + '/identities',
+        headers: {
+          Authorization: 'Bearer ' + auth0.accessToken
+        },
+        json: { provider: provider, user_id: String(providerUserId) }
+      }, function (err, response, body) {
+          if (response && response.statusCode >= 400) {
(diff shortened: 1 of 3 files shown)


gdestuynder:
Testing notes:

This is now present in Auth0-dev. @viorelaioia @rleitan this mean it is ready for testing. There is a flow doc at https://github.com/mozilla-iam/mozilla-iam/blob/master/docs/deratcheting-user-flows.md#user-logs-in-with-the-mozilla-iam-system-for-the-first-time where you have all technical details.

I realize this is fairly complex, so here's also a TLDR as well of the most common situation:

Given an existing kang@insecure.ws GitHub, FxA and LDAP account, when the user logs in as GitHub after linking is enabled, logs output:

```
    Found user_id that should be main profile used for linking, according to ratcheting logic: ad|Mozilla-LDAP-Dev|kang                                                                             
    Performing automatic profile linking: main profile: ad|Mozilla-LDAP-Dev|kang is now also main for: github|628537  
```
i.e. LDAP is selected as main profile due to "linking ratcheting" and github is linked to the LDAP account. After this there are only 2 accounts: LDAP (linked with GitHub) and FxA.

After the user also logs in with FxA or LDAP, this will happen:
```
    Found user_id that should be main profile used for linking, according to ratcheting logic: ad|Mozilla-LDAP-Dev|kang
    Performing automatic profile linking: main profile: ad|Mozilla-LDAP-Dev|kang is now also main for: oauth2|firefoxaccounts|42e3cbafe2194cc7b339c68212e06c25 
```

i.e. LDAP is again selected as main profile (again due to "linking ratcheting") and Fxa is linked to the LDAP account.

Now, there is a main LDAP account with GitHub and FxA linked.

The final user profile in Auth0 looks like:

```
user_id:  ad|Mozilla-LDAP-Dev|kang
[attributes from all 3 profiles]
identities: [ GitHub profile_data backup, FxA profile_data backup]
....
```

And the user will always look like "`ad|Mozilla-LDAP-Dev|kang` with email `kang@insecure.ws`" for all RPs.

rules/link-users-by-email-with-metadata.js
```diff
@@ -0,0 +1,117 @@
...
+                      'google-oauth2': 3,
+                      'email': 4
+                     };
+  // How old an account should be to be considered "new" to the system for linking, in seconds
+  const user_ratchet_link_delay_sec = 300;
+
+  const userApiUrl = auth0.baseUrl + '/users';
+  const userSearchApiUrl = auth0.baseUrl + '/users-by-email';
+
+  request({
```
gene1wood: Is there not anything in the `user` object that we already have (before calling `userSearchApiUrl`) which indicates that `user` is already a linked user? Couldn't we start by checking for the presence of `identities` in `user` (which over time will account for near 100% of logins) and if it's present, avoid making the call to `userSearchApiUrl`?

rules/link-users-by-email-with-metadata.js
```diff
@@ -0,0 +1,117 @@
...
+
+    // We have matches in Auth0, so we need to decide which account is going to be main
+    // Please see https://github.com/mozilla-iam/mozilla-iam/blob/master/docs/deratcheting-user-flows.md#user-logs-in-with-the-mozilla-iam-system-for-the-first-time 
+    // for detailed explanation of what happens here.
+    // NOTE: No date time check is made here as the current/new user is not yet present in Auth0 database (anymore?)
+    // thus we do not need to verify that the current profile (`user`) has a "new" `user.created_at` value. Yay!
+    for (var i = 0, len = data.length; i < len; i++) {
+      var targetUser = data[i];
+
+      var otherConnection = targetUser.identities[0].connection;
```
gene1wood: In the `identities` data structure, what is the first element in the array? Or is the array always length 1? Maybe add a comment here indicating why you're grabbing element `[0]`

rules/link-users-by-email-with-metadata.js
```diff
@@ -0,0 +1,117 @@
...
+      return u.email_verified && (u.user_id !== user.user_id);
+    });
+
+    if (data.length === 0) {
+      // Already linked or no matching email found, continue with login
+      return callback(null, user, context);
+    }
+
+    // Default to ourselves as main profile
+    var originalUser = user;
```
gene1wood: Since you're overriding the content of `originalUser` as you go, this variable shouldn't be called `originalUser` but instead should be something like `currentUser` since at the end of the loop below it should be the user with the lowest `matchOrder`

Then after you finish the loop, do something like
```
    var primaryUser = currentUser;
```

To make it clear that we've finished iteration and determined the `primaryUser`. It's confusing below when you're using a variable called `originalUser` which isn't the original user.

rules/link-users-by-email-with-metadata.js
gene1wood: You should call `otherConnection` `targetConnection` and `curConnection` `currentConnection` (assuming you go with my idea above of renaming `originalUser` to `currentUser`). This will clarify the relationship between the user and the connection

rules/link-users-by-email-with-metadata.js
```diff
@@ -0,0 +1,117 @@
...
+        console.log("Found user_id that should be main profile used for linking, according to ratcheting logic: " + targetUser.user_id);
+        originalUser = targetUser;
+      }
+    }
+
+    // Current user providers
+    const provider = user.identities[0].provider;
+    const providerUserId = user.identities[0].user_id;
+
+    if (originalUser.user_id === user.user_id) {
```
gene1wood: Ya, this would make more sense if it said
```
     if (primaryUser.user_id === user.user_id) { 
```

rules/link-users-by-email-with-metadata.js
```diff
@@ -0,0 +1,117 @@
...
+      // The profile we're trying to link is the same as the one we're login as
```
gene1wood:
```suggestion
      // The profile we're trying to link is the same as the one we're logged in as
```

rules/link-users-by-email-with-metadata.js
```diff
@@ -0,0 +1,117 @@
...
+      // This happens if the linking ratcheting logic selects the same profile and other profiles are not yet linked
+      // These will be linked the first time the user login with them instead
```
gene1wood:
```suggestion
      // These will be linked the first time the user logs in with them instead
```

rules/link-users-by-email-with-metadata.js
```diff
@@ -0,0 +1,117 @@
...
+ * @overview Link any accounts that have the same email address while merging metadata.
+ * @gallery true
+ * @category access control
+ *
+ * This rule will link any accounts that have the same email address while merging metadata.
+ * Source/Original: https://github.com/auth0/rules/blob/master/src/rules/link-users-by-email-with-metadata.js
+ *
+ */
+
+function (user, context, callback) {
```
gene1wood: We should start be defining terminology used in variables and comments because it's confusing. Something like asserting what these are
* current user
* original user
* main user
* primary user
* other user

And we should consolidate these terms in variables and comments to just what's needed.

rules/link-users-by-email-with-metadata.js
```diff
@@ -0,0 +1,117 @@
...
+    // Ignore non-verified users and current user, if present
+    data = data.filter(function (u) {
+      return u.email_verified && (u.user_id !== user.user_id);
+    });
+
+    if (data.length === 0) {
+      // Already linked or no matching email found, continue with login
+      return callback(null, user, context);
+    }
+
```
gene1wood: Don't we need to add a case here, where `data.length === 1` meaning that there is one other user with the same email address, and `data[0].identities.length === 1` meaning that the other user has not been linked and in that case, *don't* use the ratcheting logic to decide which user is primary, the other user is primary, and the current user must be linked as a child to that other identity (regardless of ratcheting logic).

This is because in the future after the transition is done and ratcheting is long gone, ratcheting should not affect which user is primary in this scenario.

This is laid out in [User logs in with second IdP for the first time](https://github.com/mozilla-iam/mozilla-iam/blob/master/docs/deratcheting-user-flows.md#user-logs-in-with-second-idp-for-the-first-time)

rules/link-users-by-email-with-metadata.js
```diff
@@ -0,0 +1,117 @@
...
+    var originalUser = user;
+
+    // We have matches in Auth0, so we need to decide which account is going to be main
+    // Please see https://github.com/mozilla-iam/mozilla-iam/blob/master/docs/deratcheting-user-flows.md#user-logs-in-with-the-mozilla-iam-system-for-the-first-time 
+    // for detailed explanation of what happens here.
+    // NOTE: No date time check is made here as the current/new user is not yet present in Auth0 database (anymore?)
+    // thus we do not need to verify that the current profile (`user`) has a "new" `user.created_at` value. Yay!
+    for (var i = 0, len = data.length; i < len; i++) {
+      var targetUser = data[i];
+
```
gene1wood: We need to add a case here where we detect if one of the `targetUser`s we're iterating over is a user that already contains multiple identities (`targetUser.identities.length > 1`). If we find a `targetUser` like this, we must set it as primary and ignore ratcheting logic to decide primary.

This case is laid out in [User logs in with third IdP and the first two are already linked](https://github.com/mozilla-iam/mozilla-iam/blob/master/docs/deratcheting-user-flows.md#user-logs-in-with-third-idp-and-the-first-two-are-already-linked)

rules/link-users-by-email-with-metadata.js
```diff
@@ -0,0 +1,117 @@
...
+    if (originalUser.user_id === user.user_id) {
+      // The profile we're trying to link is the same as the one we're login as
+      // This happens if the linking ratcheting logic selects the same profile and other profiles are not yet linked
+      // These will be linked the first time the user login with them instead
+      console.log("No automatic profile linking performed due to main profile matching current profile for: " + user.user_id);
+      return callback(null, user, context);
+    }
+
+    console.log("Performing automatic profile linking: main profile: "+originalUser.user_id+" is now also main for: " + user.user_id);
+
```
gene1wood: So in this case, we're deviating from the [User logs in with third IdP for the first time and the other two are not linked](https://github.com/mozilla-iam/mozilla-iam/blob/master/docs/deratcheting-user-flows.md#user-logs-in-with-third-idp-for-the-first-time-and-the-other-two-are-not-linked) logic and [User logs in with third IdP that they've used before and the other two are not linked](https://github.com/mozilla-iam/mozilla-iam/blob/master/docs/deratcheting-user-flows.md#user-logs-in-with-third-idp-that-theyve-used-before-and-the-other-two-are-not-linked) logic and instead of linking all non primary identities found to the primary identity, we're only linking the current identity that the user is logging in with to the primary identity.

I worry that this will extend the period of time during which users are in a partially linked state (some of their identities are linked to a primary and others are not).

Is there a reason why we only link the current user to the primary instead of linking all the non-primary users to the primary?

rules/link-users-by-email-with-metadata.js
gdestuynder: no, because while you would know if an account already has some links, you wouldnt know of the other addresses it should link to without fetching these results
on the plus side its a local call (it does not need to route over the internet) so its quite fast

rules/link-users-by-email-with-metadata.js
gdestuynder: profiles that aren't linked have only a single item, but the type of `identities` is list/array, thus, its always element 0

rules/link-users-by-email-with-metadata.js
gdestuynder: really this is just the auth0 original var, ill change it for something more clear

rules/link-users-by-email-with-metadata.js
gene1wood: What if, as we iterate over the search results, one of the returned results is a linked primary account, and as such has multiple identities?

I think this comes down to my [comment below](https://github.com/mozilla-iam/auth0-deploy/pull/236#discussion_r240766039) about the [User logs in with third IdP and the first two are already linked](https://github.com/mozilla-iam/mozilla-iam/blob/master/docs/deratcheting-user-flows.md#user-logs-in-with-third-idp-and-the-first-two-are-already-linked) case

rules/link-users-by-email-with-metadata.js
gene1wood: But in the currently logic, don't we only ever trigger a linking action (a call to establish a new link) when the current user is *not* primary?

If `user.identities.length > 1` wouldn't that indicate that the currently logged in user was primary and that we don't need to do a search at all (because even if it turned up new unliked non-primary identities, we don't do anything about that)?

If we do follow [my suggestion below](https://github.com/mozilla-iam/auth0-deploy/pull/236#discussion_r240767744) and move to a model where we always link all children to the primary, regardless of who the user is currently logging in as, in order to reduce the time that users are in a partially-linked state, then indeed, we always have to do a search, but I don't think the PR in it's current state does that.

gdestuynder: we trigger the linking action every time, as in, it looks the matches up every time (then only links if necessary) if the user is not primary. we do not know if its primary though, until we looked up all matches

rules/link-users-by-email-with-metadata.js
gdestuynder: we never link primary to itself, so it does not matter
what i should have written is: the first element (0) is always the primary user structure of the profile you're in

ie:

```
user_id: blah|blah
identities: [ {user_id: blah|blah, connection: x}, {actually linked accounts go here} ]
```
this is convenient because it saves an async lookup, though the structure is a little odd (but, that's not a structure we create, its just how it is - probably due to auth0 historical reasons)

rules/link-users-by-email-with-metadata.js
gdestuynder: i either misunderstood how this work in auth0, or this changed over time, when we wrote the doc.
auth0 in fact does not, or no longer create the profile entry at this stage of the login, so i effectively cant look up if this is a "2nd idp for the first time", but also I no longer need to. It simply isn't looked up for "linking ratcheting" since it's not in the database.

In other words:

Setup:
- you have an existing LDAP test@test.com
- you login with FxA test@test.com for the first time

Results:
- the loop finds the LDAP account only
- LDAP is selected as primary user
- FxA is linked to LDAP
user_id is ad|Mozilla-LDAP|test
identities contains : fxa-id-here

So, the logic is the same but the implementation is slightly different(but also, respecting the exact implementation is not possible)

rules/link-users-by-email-with-metadata.js
gdestuynder: thanksfully auth0 stores identities differently now, as per another comment, the primary user identity is always first in the array. I use that property.

ie:
user logins in with a github|000 user id and matches an existing profile with ad|Mozilla-LDAP|test and oauth2|blah, where the LDAP user id is the primary user.
this will link github|000 to the LDAP user id.

Note that this is because the logic never needs to ignore ratcheting, it only needs to obey the same rules every time

rules/link-users-by-email-with-metadata.js
gdestuynder: partial linked states don't really matter much in practice.
however, doing a full link increase code complexity significantly, thus, i decided to leave it that way

gdestuynder:
Filed https://github.com/mozilla-iam/mozilla-iam/issues/91 for doc fixes

PR #241, opened 2018-12-13
add logic to care for the case where an already linked account is logging     in but there is also a match in auth0 database

gdestuynder:
i first wrote this in async/await with full linking then realized auth0 rules do not support this even thus webtask does... i don't know if auth0 plans to add the support, but right now i cant access `module.exports` and thus cannot declare the function we're running in as `async` so cannot use `await`.

This means the code for this would have to loop and pass callbacks to callbacks of functions, which is extremely confusing (see the access file code for a taste, where i did not have a choice).

Due to this I found another logical way to achieve the same result by checking if `user.identities[x].profileData` exists, which means a linked profile already exist.

I reworked the comments to include how each case should function. If you find a case that you think is inaccurate we should be able to at least verify that with these test cases every time, until this gets actual TDD code (hey, I can dream right :)

I've also tested this in dev and left it in dev. I suspect that by now you have a way to go through test cases, so feel free to reproduce any i missed in dev directly if that's faster than trying to figure the code out

### DIFF
--- a/rules/link-users-by-email-with-metadata.js
+++ b/rules/link-users-by-email-with-metadata.js
@@ -7,6 +7,9 @@
  * This rule will link any accounts that have the same email address while merging metadata.
  * Source/Original: https://github.com/auth0/rules/blob/master/src/rules/link-users-by-email-with-metadata.js
  *
+ * Please see https://github.com/mozilla-iam/mozilla-iam/blob/master/docs/deratcheting-user-flows.md#user-logs-in-with-the-mozilla-iam-system-for-the-first-time
+ * for detailed explanation of what happens here.
+ *
  */
 
 function (user, context, callback) {
@@ -32,11 +35,6 @@ function (user, context, callback) {
   const userApiUrl = auth0.baseUrl + '/users';
   const userSearchApiUrl = auth0.baseUrl + '/users-by-email';
 
- 
-  // primaryUser : The user profile that all other user profiles will be linked
-  //               to as children
-  // targetUser : The iterator that we use as we iterate over the search results
-
   request({
    url: userSearchApiUrl,
    headers: {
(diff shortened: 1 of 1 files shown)


rules/link-users-by-email-with-metadata.js
```diff
@@ -61,37 +59,57 @@ function (user, context, callback) {
...
+    // FxA logins: CASE 3 is hit (LDAP primary, Fxa linked)
+    // FxA logins again: No change, due to this case
+    // Eventually GitHub login and hits CASE 2
+    if (data.length === 1 && user.identities && user.identities.length > 1) {
+      primaryUser = user;
+
+    // CASE 2:
+    // If we only find a single account in the Auth0 database, we do not apply ratcheting logic as this means
+    // 1) `user` is a new user not in the database
+    // 2) `targetUser` is already linked to something, or is a single unlinked account and thus should be
```
gene1wood: Case 2, as you lay out correctly in the comment
a. if the user is logging in with an unlinked identity and there is another identity in Auth0 which is a linked account
b. if the user is logging in with an unlinked identity and there is another identity in Auth0 which is not a linked account

The logic here treats 2a and 2b the same and sets the primary to whatever the *other* identity is.

I think however that 2b should instead use ratcheting. Here's an example to illustrate it

1. Two identities exist in Auth0 with the same email address, a GitHub and a passwordless
2. We enable account linking
3. User logs in with GitHub
4. We match against Case 2 because
  * The search returns `data.length` of 1
  * `user.identities.length` is 0 because the GitHub account is not linked
5. We set `primaryUser` to `targetUser`. `targetUser` is passwordless
6. The rule makes a linking call setting passwordless to primary and GitHub to non-primary
7. The user continues onto the RP with the passwordless `user_id` after linking happens
8. The user is unable to reach their now orphaned account on the RP because the account previously used the GitHub `user_id` due to ratcheting.

I think we need to examine `targetUser` to determine if it has an `identities.length > 1`. If it does, then we set `primaryUser = targetUser`. If it does not, then we use ratcheting to decide if `user` or `targetUser` should be set as primary.

Note, that this nuance is not reflected in our design doc : https://github.com/mozilla-iam/mozilla-iam/blob/master/docs/deratcheting-user-flows.md#user-logs-in-with-second-idp-for-the-first-time

Either I'm overlooking something here or this is something we didn't account for in the initial design doc.

gdestuynder: I'm thinking about this:

```
diff --git a/rules/link-users-by-email-with-metadata.js b/rules/link-users-by-email-with-metadata.js
index 3c672bf..a16f30b 100644
--- a/rules/link-users-by-email-with-metadata.js
+++ b/rules/link-users-by-email-with-metadata.js
@@ -84,11 +84,12 @@ function (user, context, callback) {
     // Example test case A: LDAP or a linked LDAP account exists as well as Github, all with kang@insecure.ws `user.email`
     // LDAP logins: Github is set as target here and will therefore be linked to LDAP or linked LDAP as primary
     // Example test case B: same setup but
-    // GitHub logins: CASE 1 is hit if LDAP is already linked
-    //                CASE 3 is hit if neither accounts are linked but GitHub already existed
+    // GitHub logins: CASE 1 is hit if GitHub is already linked
+    //                CASE 3 is hit if neither accounts are linked but GitHub already exist
+    //                CASE 3 is hit if LDAP is already linked (because LDAP, i.e. targetUser.identities.length > 1)
     //                CASE 2 (THIS CASE) is hit if GitHub is a new account and it will be the primary account, LDAP will
-    //                be linked to it
-    } else if (data.length === 1) {
+    //                be linked to it as LDAP is not already linked
+    } else if (data.length === 1 && targetUser.identities && targetUser.identities.length <= 1) {
       primaryUser = targetUser;
 
     // CASE 3:
```

gdestuynder:
This made me think of another new corner case:

- user has a linked account (GitHub + linked FxA for example), email test@test.com
- user goes to Github and change email to hi@hi.com, then logs in at least once
- user now logs in with passwordless email hi@hi.com

when the user logs in, the system would try to link passwordless to github (email hi@hi.com), which is itself linked to FxA and causes a "conflict". I believe auth0 then selects hi@hi.com as the primary email and retains the FxA profile link.

This is what my comment also refers to here https://github.com/mozilla-iam/mozilla-iam/issues/91

maybe we're just fine with that. i dont have a good way to "fix it" other than what the comment says

rules/link-users-by-email-with-metadata.js
gdestuynder: ive added this in the pr

PR #252, opened 2019-02-07
remove groups for AAL=LOW users

gdestuynder:


gene1wood:
Ya, this is an important part of the design of Login V2

Any RP that wants to be set to ALL=LOW must satisfy one of these two cases

* The RP doesn't handle sensitive data and so they don't have any need for users to provide a higher level of identity assurance than LOW
* The RP does handle sensitive data but wants to be able to surface some site functionality (which does not put sensitive data at risk) to users who log in with a LOW level of identity assurance. An example of this would be an RP that has and administrative web UI and an ability for users to comment on posts. The RP would want to require a high level of identity assurance when users access admin functions but also to allow LOW level identity assurance users to be able to comment on posts.

For the second case, before we can change them to be set to AAL=LOW, the RP must

* illustrate to the IAM team (in written form, for example in a bug/issue) the use case of what subset of their site features they intend to make available to AAL=LOW users
* access to the code that is governing this differentiation so that the IAM team can review the code quality and logic
* agree that if they intend any functional change to their site behavior that will surface additional functionalities to AAL=LOW users that they inform the IAM team before they make the change to give the IAM team an opportunity to weigh in. This notification needs to be written (e.g. a bug/issue) and needs to show that an RRA has been done on the service and the data classification of the data that this new functionality would surface to AAL=LOW users

Because of this, we should not make some technical change in Login V2 like what's present in this PR. If there is a problem with the design described above, let's meet and come to an agreement about a change to the design (instead of implementing technical shims like what's in this PR)

gdestuynder:
@hmitsch @akatsoulas @viorelaioia

hmitsch:
@gene1wood and @gdestuynder, I agree with @gene1wood's comment above (https://github.com/mozilla-iam/auth0-deploy/pull/252#issuecomment-461523099) and confirm that we will proceed to handle "Case 2 AAL=LOW" desirements as outlined.

-Henrik

PR #258, opened 2019-02-14
cleanup + fixes

gdestuynder:
- cleanup case to have CLEAR if conditions between data.length and identities (no change in functionality)
- drop unused error case (one case is no longer triggered, nor needs to be since there are no other options)
- make error cases send back useful error messages (instead of `body` - didnt see it last PR)
- add support for ratchet time delay (where did it go?) as this is needed - this fixes an actual bug (and why i started looking at this yet again) where you have 2 accounts in the database but we only use 1 or them, even if they were present pre-deratcheting. I'm not sure why the variable is still there but the code was gone. Maybe it never got implemented fully and we never noticed.

this is in dev and pre-tested with some cases, but requires all cases to run again..

### DIFF
--- a/rules/link-users-by-email-with-metadata.js
+++ b/rules/link-users-by-email-with-metadata.js
@@ -46,19 +46,23 @@ function (user, context, callback) {
   },
   function (err, response, body) {
     if (err) return callback(err);
-    if (response.statusCode !== 200) return callback(new Error(body));
+    if (response.statusCode !== 200) return callback(new Error("API Call failed: "+body));
 
     var data = JSON.parse(body);
     // Ignore non-verified users and current user, if present
+    // The user_ratchet_link_delay_sec dictate if we consider this user to be present pre or post-ratcheting
     data = data.filter(function (u) {
-      return u.email_verified && (u.user_id !== user.user_id);
+      var ret = u.email_verified;
+      var creation_time_distance = Date.now() - Date.parse(u.created_at);
+      if (creation_time_distance > user_ratchet_link_delay_sec) {
+        ret = ret &&  (u.user_id !== user.user_id);
+      }
+      return ret;
     });
 
 
     // @primaryUser JSON The user profile that all other user profiles will be linked to as children
-    // @targetUser JSON The iterator that we use as we iterate over the search results, the to-be primaryUser
-    //
-    // Default to ourselves as the primaryUser and the targetUser as the first search result from Auth0 DB
+    // @targetUser JSON The first search result from Auth0 DB
     var primaryUser = user;
     var targetUser = data[0] || undefined ;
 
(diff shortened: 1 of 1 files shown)


viorelaioia:
@gdestuynder . this issue is still happening. 

STR:
1. Delete github 2fa and fxa 2fa identities from auth0
2. Turn the linking rule off
3. Login with github 2fa in testrp
4. Login with fxa 2fa in testrp
5. Turn the linking rule on
6. Login with fxa 2fa

Expected:
fxa user_id with 2 identities is returned (fxa 2fa and github 2fa identities)

Actual:
github user_id with 2 identities is returned (github 2fa and fxa 2fa identities)

gdestuynder:
because the ratchet time delay is now enabled, you will only get a fxa user id if the login after the delay has elapsed (300s)

this is what decides if the account is "new" or not. sorry, first comment wasn't super clear that this affects testing

rules/link-users-by-email-with-metadata.js
```diff
@@ -68,43 +72,48 @@ function (user, context, callback) {
...
-    // Example test case B: same setup but
-    // GitHub logins: CASE 1 is hit if GitHub is already linked
-    //                CASE 3 is hit if neither accounts are linked but GitHub already exist
-    //                CASE 3 is hit if LDAP is already linked (because LDAP, i.e. targetUser.identities.length > 1)
-    //                CASE 2 (THIS CASE) is hit if GitHub is a new account and it will be the primary account, LDAP will
-    //                be linked to it as LDAP is not already linked
-    } else if (data.length === 1 && (user.identities && user.identities.length === 1)
-               && (targetUser.identities && targetUser.identities.length >= 1)) {
-      primaryUser = targetUser;
+    // We have a single match in the auth0 database
```
gene1wood: It might be worth calling out in this comment that it's a single match *other than the one they just logged in with*

rules/link-users-by-email-with-metadata.js
```diff
@@ -125,16 +135,7 @@ function (user, context, callback) {
...
-    // ERROR CASE
-    // If user has a single account in auth0 database other than the one they just logged in with (data.length === 1),
-    // AND they are logging in with an account which is linked to one other account (user.identities.length === 1),
-    // AND that one other account has an empty identities list (targetUser.identities.length === 0)
-    // Error out because targetuser.identities.length should never be 0, it should always contain at least the identity
-    // of the profile that it lives within
-    } else {
-      return callback(new Error(body));
-    } // End of CASEs
+    }
```
gene1wood: With the final error case removed where `targetUser.identities.length === 0`, what happens now if `targetUser.identities.length === 0`? Or is this an impossible state that we need not check for?

gdestuynder: impossible state

rules/link-users-by-email-with-metadata.js
gdestuynder: i added a comment about the filtering function (same reason, just that it can be the one they just logged in with, depending on the filtering results)

PR #267, opened 2019-04-25
adding Zoom to hrdata rule to get costcenter mapping

jdow:


gdestuynder:
@jdow what does zoom uses it for?

jdow:
They want to map the department field. I'm not sure what in particular Zoom uses it for.

jdow:
I heard back. It's to automatically assign roles to specific teams, which grant permissions

PR #268, opened 2019-05-03
Fix LDAP group handling by using unique ldap_groups root attribute

jdow:


gdestuynder:
Since Viorella and Roxana will test this I'm merging it into dev (this wont be in production yet)

PR #273, opened 2019-05-21
adding SAML rule for LGTM for client specific settings

jdow:


gdestuynder:
please add a reviewer so that this does not get missed instead of directly deploying to production.

PR #281, opened 2019-09-05
Support wildcards in AWS-Federated-AMR role picker

april:
With a little bit of documentation on how to use it.

I would probably restructure this code if I had the chance to use something like `webpack` or `browserify` so that we could actually use modules, but it's a pretty big project.


rules/AWS-Federated-AMR.js
```diff
@@ -1,6 +1,26 @@
 /*jshint esversion: 6 */
 
 function (user, context, callback) {
+  // modules/group-intersection.js
+  const groupIntersection = (groups, filter) => {
+    const overlap = new Set();
+
+    const filters = filter.map(i => {
+      return new RegExp(i.replace(/\./g, '\.').replace(/\?/g, '.').replace(/\*/g, '.*'));
```
gene1wood: Took me a while to understand what this does, maybe add some comments like
* First converts all `.` characters to `\.` to escape them for regex
* Next converts all `?` characters to `.`
* Finally converts all `*` characters to `.*`

gene1wood: Also, I wonder if there's a more complete method to escape a string for use in regexp (more complete than converting `.` to `\.`). I would worry that a group name with another regex character would cause some chaos here, for example groups with names like
* `HR (temp)`
* `HR [temp]`
* `HR June+July`
* `ISC^2`

It looks like Auth0 includes [lodash](https://auth0-extensions.github.io/canirequire/#lodash) which [may have a solution](https://auth0-extensions.github.io/canirequire/#lodash) or just spin our own [more complete conversion](https://stackoverflow.com/a/30851002/168874)

april: The "may have a solution" links to the same place as "lodash". Was it supposed to link to a function?

april:
Okay, I pulled in a bunch of `lodash.escapeRegExp`, minus `.` and `?`, and added a test to cover it.

How does this look?

rules/AWS-Federated-AMR.js
gene1wood: Whoops, ya I mispasted the link to this : https://stackoverflow.com/a/29700268/168874

I've updated the comment.

PR #279, opened 2019-08-15
Add support for new S3 hosted group role map file

gene1wood:
* Define style standard as Google style 
* Add documentation on known issues

PR #286, opened 2019-09-13
Bring production up to date with master

gene1wood:
This will bring the production branch up to date with what's live in production. It will also apply two changes that are live in dev, and present in master but not yet live in production.

* Fix a typo in a comment in `temporary-LDAP-re-reintegration` in #280
* Significant rewrite of `AWS-Federated-AMR` form #279 and #281


buildspec.yml
```diff
@@ -0,0 +1,14 @@
+version: 0.2
+
+env:
+  parameter-store:
+    CLIENTID: "/iam/auth0-deploy/stage/AUTH0_RULE_DEPLOY_CLIENTID"
```
gdestuynder: will this be a problem for prod? the path seem harcoded to `/stage/`

rules/AWS-Federated-AMR.js
```diff
@@ -0,0 +1,129 @@
...
+          overlap.add(group);
+          break;
+        }
+      }
+    });
+
+    return [...overlap];
+  };
+
+  const WHITELIST = [
```
gdestuynder: for performance reasons you could-should move this to the top and return the callback immediately if you're not in the whitelist

this saves use from initializing all functions during login time, which is not completely negligible

buildspec.yml
gene1wood: Ya, this PR doesn't accommodate adding in CI for production.

rules/AWS-Federated-AMR.js
gene1wood: As we've not tested that in dev yet. Should we
* scrap this PR, make this change in dev, test, and make a new PR to prod
* defer this perf improvement, work on merging this now

Also, doesn't that just define the function but it doesn't get called unless the whitelist conditions are met? (cc @april )

buildspec.yml
gene1wood: @kang do you think it's ok to not solve the production CI question in this PR, but instead get this merged, then develop prod CI subsequently?

gdestuynder: yup thats fine, just make sure prs to prod wont actually deploy the prod branch to dev in the mean time

rules/AWS-Federated-AMR.js
gdestuynder: yes, this should go in dev first

gene1wood: Ok, maybe I can save this PR but holding on it here, making the change in dev, testing it there and then merging that commit into this PRs production branch.

@april would you be able today to try Guillaume's suggested change in dev today (with a PR to master)?

april: JavaScript will hoist the function to the top of its scope regardless of where they are declared inside the scope.

I'm not sure that V8's performance will markedly differ by moving the function declaration underneath this (due to how it JIT's). I can do this, but I'm not certain it will make any difference and will make the code less legible.

gdestuynder: you can easily check that by adding a syntax error up top and the login will fail, while if you add it after any exit loop it wont
using wt-cli you can time the initializing time


webtask uses node's sandboxing function with some modifications (`vm4` iirc)

its not a huge deal but it does not cost much - the most important though is the exit loop so that the code block is never entered for RPs that are not concerned by it - yes, not a huge deal if the code is perfect, but it never is

april: a syntax error inside the function?

gdestuynder: yup

Makefile
```diff
@@ -0,0 +1,29 @@
...
+	pip install -r requirements.txt
+
+
+deploy-local:
+	# Requires a credentials.json file to be present and valid
+	# Useful for local tests
+	uploader_rules.py -r rules
+
+deploy:
+	@echo "Deploying to $(URI)"
```
gdestuynder: we should run April's test here (make a target called `test`  then make deploy depend on it)

gene1wood: Good idea I'll track it here #290

rules/AWS-Federated-AMR.js
gene1wood: This is now moved into the conditional section dependent on the clientID

gene1wood:
This was deployed live at 7:35am PDT

PR #291, opened 2019-09-18
Fix incorrect rule variable names

gene1wood:
This check doesn't match the actual `configuration` variable names used.

You can see [the correct variable names here](https://github.com/mozilla-iam/auth0-deploy/blob/3578864a4140df364857901b955303026604722f/rules/AWS-Federated-AMR.js#L10-L13)

Once this is merged and deployed, we need to go in and delete the 3 no longer used variables from auth0 dev and prod

### DIFF
--- a/rules/AWS-Federated-AMR.js
+++ b/rules/AWS-Federated-AMR.js
@@ -44,11 +44,11 @@ function (user, context, callback) {
       return [...overlap];
     };
 
-    if (!("group_role_map_s3_bucket" in configuration) ||
-        !("group_role_map_aws_access_key_id" in configuration) ||
-        !("group_role_map_aws_secret_key" in configuration)) {
+    if (!("auth0_aws_assests_s3_bucket" in configuration) ||
+        !("auth0_aws_assests_access_key_id" in configuration) ||
+        !("auth0_aws_assests_access_secret_key" in configuration)) {
       console.log("Enriching id_token with amr for AWS Federated CLI");
-      throw new Error("missing configuration values");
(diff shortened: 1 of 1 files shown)


rules/AWS-Federated-AMR.js
```diff
@@ -44,11 +44,11 @@ function (user, context, callback) {
...
 
-    if (!("group_role_map_s3_bucket" in configuration) ||
-        !("group_role_map_aws_access_key_id" in configuration) ||
-        !("group_role_map_aws_secret_key" in configuration)) {
+    if (!("auth0_aws_assests_s3_bucket" in configuration) ||
+        !("auth0_aws_assests_access_key_id" in configuration) ||
+        !("auth0_aws_assests_access_secret_key" in configuration)) {
       console.log("Enriching id_token with amr for AWS Federated CLI");
-      throw new Error("missing configuration values");
+      throw new Error("Missing Auth0 AWS Federated AMR rule configuration values");
```
gdestuynder: quick note, this will cause an error screen on auth0 and the message is hidden by default. I didnt catch that before, but just making sure you know this - i assume this one error isn't likely to occur since variables are always set

gene1wood: The message actually propagates all the way up to the user. I encountered this error as a user and wanted to make it more explicit what it meant since it took me a while to figure out what "missing configuration values" meant.

It's conveyed back to the user in query parameters to the `redirect_uri`.

```
error=access_denied&error_description=missing%20configuration%20values
```

> i assume this one error isn't likely to occur since variables are always set

ya

PR #292, opened 2019-09-18
Deploy AWS rule fix to production

gene1wood:
This deploys #291 into production.

I've deployed #291 to dev and removed the old variables to confirm logins for AWS still work and they do.


gene1wood:
I'll plan to merge and deploy this tomorrow morning

gene1wood:
This was deployed in production at 2019-09-24 07:11:03 PDT

I've manually deleted the old environment variables that this PR relates to that are no longer in use and confirmed AWS login works.

PR #298, opened 2019-11-06
Attempt to use the apps.yml's app.url as redirect_uri when issuing error msgs first. Fallback to sso.mozilla.org otherwise.

gdestuynder:
This means errors no longer carry the actual redirect_uri as this rarely
works with RPs (they often don't know what to do when you're sent back
to their redirect_uri without any login data, which leads to bad UX)

This PR will change the link for the `Go Back` button from what it is now, which is the RP's redirect URI, to instead link to the `app.url` that comes from the `apps.yml` file which is the homepage of the RP. In the case where there is no `app.url` for the RP, the `Go Back` button will link to the SSO Dashboard


gdestuynder:
this blocks https://jira.mozilla.com/browse/IAM-329

PR #302, opened 2019-11-07
Bypass CIS fixup for requests with scopes of "openid" or "email"

### LINKED ISSUE #301: How to request scopes that return only `openid` and `email` without profile data
This issue stems from #269 

How can we request the scopes of `openid` and `email` and get only those claims. We need this for the AWS federated CLI. Currently the logic in #269 causes any scope request other than `openid` to send along custom claims.

Can we change the logic to send custom claims when a user requests `profile` scope?

gene1wood:
This is a workaround for #301. We should make a proper fix after notifying RPs
by checking for the "profile" scope instead.

Fixes #301

### DIFF
--- a/rules/CIS-Claims-fixups.js
+++ b/rules/CIS-Claims-fixups.js
@@ -1,6 +1,7 @@
 function (user, context, callback) {
   var namespace = 'https://sso.mozilla.com/claim/';
   var whitelist = ['']; // claim whitelist
+  let isSetsEqual = (a, b) => a.size === b.size && [...a].every(value => b.has(value));
 
   // If you're not OIDC Conformant, stop right there as this INCREASES the profile size
   // significantly! The metadata is set by us, manually.
@@ -14,9 +15,11 @@ function (user, context, callback) {
   }
 
 
-  // If the only scope requested is openid, do not overload with custom claims
-  if (context.request.query.scope === "openid") {
-    console.log('Client '+context.clientID+' only requested scope:openid, not adding custom claims');
+  // If the only scopes requested are openid or email, do not overload with custom claims
+  let scopes = new Set(context.request.query.scope.split(' '));
+  let whitelisted_scopes = new Set(["openid", "email"]);
+  if (isSetsEqual(scopes, whitelisted_scopes)) {
+    console.log('Client '+context.clientID+' only requested scope:openid or scope:email, not adding custom claims');
     return callback(null, user, context);
   }
 
--- a/rules/CIS-Claims-fixups.js
+++ b/rules/CIS-Claims-fixups.js
@@ -1,7 +1,7 @@
 function (user, context, callback) {
+  var _ = require('lodash');
   var namespace = 'https://sso.mozilla.com/claim/';
   var whitelist = ['']; // claim whitelist
-  let isSetsEqual = (a, b) => a.size === b.size && [...a].every(value => b.has(value));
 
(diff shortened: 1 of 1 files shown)


rules/CIS-Claims-fixups.js
```diff
@@ -1,6 +1,7 @@
 function (user, context, callback) {
   var namespace = 'https://sso.mozilla.com/claim/';
   var whitelist = ['']; // claim whitelist
+  let isSetsEqual = (a, b) => a.size === b.size && [...a].every(value => b.has(value));
```
gdestuynder: wondering if @april knows a prettier way to do this (like `Array.from(a).every(b);` if that works in this node)

rules/CIS-Claims-fixups.js
```diff
@@ -14,9 +15,11 @@ function (user, context, callback) {
   }
 
 
-  // If the only scope requested is openid, do not overload with custom claims
-  if (context.request.query.scope === "openid") {
-    console.log('Client '+context.clientID+' only requested scope:openid, not adding custom claims');
+  // If the only scopes requested are openid or email, do not overload with custom claims
+  let scopes = new Set(context.request.query.scope.split(' '));
+  let whitelisted_scopes = new Set(["openid", "email"]);
```
gdestuynder: @gene1wood any reason for not doing the opposite kind of logic?
ie:
- if `profile` is in add custom claims
- if one scope startswith ('https') add custom claims
- else dont add extra custom claims

gene1wood: Reason was I figured we could make this change without notification to all RPs and waiting some window of time before making the change. My assumption was from #269, that changing to requiring the presence of `profile` scope in order to do CIS fixups, would require notifying RPs first.

This PR is a workaround in advance of the "right" fix of checking for `profile`

rules/CIS-Claims-fixups.js
april: Javascript is pretty barebones when it comes to set support. This is a fine way to do it, but Auth0 gives you access to `lodash` so I would probably just import `isEqual` from `lodash` and call it a day.

rules/CIS-Claims-fixups.js
```diff
@@ -14,9 +15,11 @@ function (user, context, callback) {
...
+  if (isSetsEqual(scopes, whitelisted_scopes)) {
+    console.log('Client '+context.clientID+' only requested scope:openid or scope:email, not adding custom claims');
```
april: If you use `_.difference()`, you can log what things were requested but not in the whitelist.

rules/CIS-Claims-fixups.js
gene1wood: I've changed this to use lodash in 1d2d3e215741e97cfd2756a8ce28a14ae45c6402

rules/CIS-Claims-fixups.js
gene1wood: I've changed this to lodash in 1d2d3e215741e97cfd2756a8ce28a14ae45c6402 and fixed the bad scope comparison logic

rules/CIS-Claims-fixups.js
gene1wood: I've changed the logic to what @gdestuynder suggests above in b0cea30bfb7e9ff00f057408eb8985b396de348e

gene1wood:
This has been deployed to dev but needs merging to production and deployment to prod

PR #305, opened 2019-11-19
Merge master to production

jdow:


gene1wood:
@jdow Remind me, this came up last week or the week before in talking with @gdestuynder . Dev (`master`) is ahead of prod (`production`) at the moment I'm assuming and as would make sense.

@gdestuynder was asking about commits present in this PR (I think) that weren't present in `master` or maybe it was `production`, I can't remember.

What do you think the next steps are on this PR and on getting production in line with dev?

PR #313, opened 2020-02-27
Update force-ldap-logins-over-ldap.js

jdow:
I'd like to remove mozilla.org, mozilla.net and readitlater.com domains from this rule, because these are not LDAP managed domains (and never have been), and are not ever issued to staff for any reason.

PR #314, opened 2020-02-28
First step in rewriting every rule and giving things tests

april:
The code was... a little rough. I've done my best to make things as clear and understandable as possible.

There are tests for these (and they all work locally), but it should probably get deployed to dev for a little bit before going to production. Subsumes #313.


april:
Actually, lemme cancel this and I'm going to push a new PR with actual test stuff in it.

PR #324, opened 2020-04-28
Fix test due to change by @jdow

april:


april:
Should be a quick review, @gene1wood.  :)

Issue #325, opened 2020-04-30
Too many SAML configuration mapping rules


april:
There are currently 11 SAML configuration rules, which makes for a very long ruleset in auth0.

It would be great to consolidate a bunch of these down into a single (or at least fewer) rules.

PR #326, opened 2020-05-06
Start consolidating SAML rules

april:
Start consolidating and fixing #325. In theory this should be faster and a lot messier. I only did two rules to begin with, along with their respective tests.


rules/SAML-configuration-mapping.js
```diff
@@ -0,0 +1,33 @@
...
+
+      context.samlConfiguration.mappings = {
+        'Company Name': 'company_name',
+        'emailAddress': 'email',
+        'name':         'name',
+      };
+
+      break;
+    case 'thinksmart':
+      context.samlConfiguration.mappings = {
```
gene1wood, line 23: Is it intentional that this overwrites the contents of `context.samlConfiguration.mappings` whereas the existing rule uses [`extend`](https://www.npmjs.com/package/extend)

rules/SAML-configuration-mapping.js
```diff
@@ -0,0 +1,33 @@
...
+
+  const client = CLIENTS[context.clientID];
+
+  switch(client) {
+    case undefined:
+      break;
+    case 'sage-intacct':
+      user.company_name = 'MOZ Corp';
+
+      context.samlConfiguration.mappings = {
```
gene1wood, line 15: We should probable add this in case the object is missing from `context`
```suggestion
      context.samlConfiguration = context.samlConfiguration || {};
      context.samlConfiguration.mappings = {
```

rules/SAML-configuration-mapping.js
april: As far as I know, there is only a 1:1 mapping of `context.clientID` to `samlConfiguration`, so there is no reason to call `extend()` at all.

april: Also, none of the other rules use extend, only this one. So I would be shocked if it were necessary.

rules/SAML-configuration-mapping.js
april: We can, although none of the other rules do this. I'm pretty sure it sets `context.samlConfiguration` automatically.  :)

april: Okay, made another push that cleans things up a bit and has a protective measure in there for you.  :)

PR #327, opened 2020-05-14
Remove rules for mozilla-iam AWS SAML integration

gene1wood:
This is deprecated in favor of the OIDC integration
See IAM-543


april:
Happy to see it, this rule is a mess too.  :)

PR #329, opened 2020-05-20
Revamp HRIS-is-staff (previously temporary-hris-connector)

april:
Rename temporary-hris-connector.js to HRIS-is-staff, rewrite good chunks of it, add tests.


rules/HRIS-is-staff.js
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  const isStaff = user._HRData !== undefined && user._HRData.cost_center !== undefined;
+  user.app_metadata = user.app_metadata || {};
+  user.app_metadata.groups = user.app_metadata.groups || [];
+  user.groups = user.groups || [];
+
+  // add `hris_is_staff` to groups if not there and user is staff
+  if (isStaff &&
+      !user.app_metadata.groups.includes('hris_is_staff') &&
```
gene1wood, line 30: Will there ever be a case where `user.app_metadata.groups` has `hris_is_staff` but `user.groups` does not, or vice versa? Because if so that case isn't handled

april: I don't think so, but I'll go ahead and add it.

PR #331, opened 2020-05-21
Add Stripe subplat saml mapping support

gene1wood:
https://bugzilla.mozilla.org/show_bug.cgi?id=1637117


rules/SAML-configuration-mapping.js
```diff
@@ -33,6 +34,33 @@ function (user, context, callback) {
       };
 
       break;
+    case 'stripe-subplat':
```
april, line 37: This code path has no tests and also does not compile. It looks like it got pushed into auth-dev, and that is causing the `SAML-configuration-mapping.js` rule to fail.

rules/SAML-configuration-mapping.js
```diff
@@ -33,6 +34,33 @@ function (user, context, callback) {
...
+        'stripe_subplat_developer': [
+          {'role': 'developer', 'account': 'acct_1EJOaaJNcmPzuWtR'}],
+        'stripe_subplat_supportsp': [
+          {'role': 'support_specialist', 'account': 'acct_1EJOaaJNcmPzuWtR'}],
+        'stripe_subplat_analyst ': [
+          {'role': 'analyst', 'account': 'acct_1EJOaaJNcmPzuWtR'}],
+        'stripe_subplat_viewonly ': [
+          {'role': 'view_only', 'account': 'acct_1EJOaaJNcmPzuWtR'}]
+      };
+      context.samlConfiguration = context.samlConfiguration || {};
```
april, line 53: There is no need to do this, as it is done in line 16.

rules/SAML-configuration-mapping.js
```diff
@@ -33,6 +34,33 @@ function (user, context, callback) {
...
+      context.samlConfiguration.mappings = context.samlConfiguration.mappings || {};
+      Object.keys(groupToStripeRoleMap).forEach((groupName) => {
+        if (user.hasOwnProperty('groups') && user.groups.includes(groupName)) {
+          Object.keys(groupToStripeRoleMap[groupName]).forEach((roleInfo) => {
+            user.app_metadata[roleInfo['account']] = roleInfo['role'];
```
april, line 58: Use dot syntax here, e.g. `roleInfo.account` and `roleInfo.role`.

rules/SAML-configuration-mapping.js
```diff
@@ -33,6 +34,33 @@ function (user, context, callback) {
...
+          });
+          context.samlConfiguration.mappings['Stripe-Role-' + roleInfo['account']] = 'app_metadata.' + roleInfo['account'];
```
april, line 60: Use templates here, and dot syntax.

rules/SAML-configuration-mapping.js
```diff
@@ -33,6 +34,33 @@ function (user, context, callback) {
...
+        }
+      });
+    }
```
april, line 63: I think this is too many brackets, causing the return to be stranded.

rules/SAML-configuration-mapping.js
```diff
@@ -33,6 +34,33 @@ function (user, context, callback) {
       };
 
       break;
+    case 'stripe-subplat':
+      // https://bugzilla.mozilla.org/show_bug.cgi?id=1637117
+      var groupToStripeRoleMap = {
+        //  LDAP group name
+        'stripe_subplat_admin': [
+          //   stripe_role_name           stripe_account_id
+          {'role': 'admin', 'account': 'acct_1EJOaaJNcmPzuWtR'}],
```
april, line 43: I would say to format these either as normal or to condense it down to five lines. This is pretty hard to read.

rules/SAML-configuration-mapping.js
```diff
@@ -33,6 +34,33 @@ function (user, context, callback) {
       };
 
       break;
+    case 'stripe-subplat':
+      // https://bugzilla.mozilla.org/show_bug.cgi?id=1637117
+      var groupToStripeRoleMap = {
```
april, line 39: Oh, and `const`, not `var`.  :)

rules/SAML-configuration-mapping.js
gene1wood: Aha, thank you. Ok I've fixed this and added tests.

rules/SAML-configuration-mapping.js
gene1wood: Ah good point, I've removed it.

rules/SAML-configuration-mapping.js
gene1wood: I think we want `var` here to follow the Auth0 style : https://github.com/mozilla-iam/auth0-deploy#style

rules/SAML-configuration-mapping.js
gene1wood: Yes, got it. I've fixed this and added the missing `break;`

rules/SAML-configuration-mapping.js
gene1wood: Sure thing

rules/SAML-configuration-mapping.js
gene1wood: Good idea, I've changed it to 5 lines.

rules/SAML-configuration-mapping.js
gene1wood: Sure thing

rules/SAML-configuration-mapping.js
april: We should change our style then. :)

Const and let are generally how things should be done, and the rest of this code (as well as all of my new code) uses const/let.

april: To add, our documentation says:

> The primary goal is to follow the style of the Auth0 example rules.

Having looked at the auth0 example rules, it seems like they have all been updated to use `const` and `let`.  :)

gene1wood: Ok, I opened #333 to capture that. I'll leave this as is and we can clean this rule up with all the other rules and the documentation.

april:
This rule was already cleaned up, by leaving it this way you’re making it
so it now needs to be cleaned up.  :P

On May 27, 2020, at 10:34 AM, Gene Wood <notifications@github.com> wrote:

﻿

*@gene1wood* commented on this pull request.
------------------------------

In rules/SAML-configuration-mapping.js
<https://github.com/mozilla-iam/auth0-deploy/pull/331#discussion_r431236817>
:

> @@ -33,6 +34,33 @@ function (user, context, callback) {
       };

       break;
+    case 'stripe-subplat':
+      // https://bugzilla.mozilla.org/show_bug.cgi?id=1637117
+      var groupToStripeRoleMap = {

Ok, I opened #333 <https://github.com/mozilla-iam/auth0-deploy/issues/333>
to capture that. I'll leave this as is and we can clean this rule up with
all the other rules and the documentation.

—
You are receiving this because your review was requested.
Reply to this email directly, view it on GitHub
<https://github.com/mozilla-iam/auth0-deploy/pull/331#discussion_r431236817>,
or unsubscribe

gene1wood:
Ok, I've changed `var` to `const`

PR #334, opened 2020-06-03
Add direct creation of users in CIS from Auth0

april:


april:
Note that this still needs tests, like... so, so many tests.

PR #338, opened 2020-07-21
Add different ikey and skey for login

april:
As well as fix up a lot of code and add more tests.

### DIFF
--- a/rules/duosecurity.js
+++ b/rules/duosecurity.js
@@ -1,5 +1,10 @@
 function (user, context, callback) {
-  var WHITELIST = [
+  const CLIENTS_WHICH_ALLOW_ENROLLMENT = [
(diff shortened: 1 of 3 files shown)


april:
cc: @jdow, @rtucker-mozilla, @cshields

To verify that this matches up against their design vision. I've already added the configuration variables to test, so we would simply need to merge the rule change in and it would start working.

rtucker-mozilla:
I  happily defer to your expertise @april

april:
Should we still move forward on this project? If so, I can get this merged provided @gene1wood is able to provide review.

gene1wood (commented):
I'm unsure how this fits in with the IAM teams desires here in 2024

dividehex:
This is no longer needed.

PR #341, opened 2020-08-14
Add _email as an attribute to all users

april:
Many SAML apps need email/id in multiple attributes, this just makes them always available in the user object.

### DIFF
--- a/rules/SAML-configuration-mapping.js
+++ b/rules/SAML-configuration-mapping.js
@@ -6,6 +6,10 @@ function (user, context, callback) {
   };
   const client = CLIENTS[context.clientID];
 
+  // many SAML implementations require your id (email address) twice, but SAML configurations
+  // inside auth0 can only map from an attribute once
+  user._email = user.email;
+
   // if it's not a client that uses SAML, exit immediately
   if (!client) {
     return callback(null, user, context);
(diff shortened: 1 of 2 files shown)


gene1wood:
I've asked Auth0 in this support ticket how they intend customers to use the SAML2 Web App addon and be able to map a single Auth0 field to multiple SAML fields

https://support.auth0.com/tickets/00464575

> When enabling a SAML RP using the Addon: SAML2 Web App in Auth0, we need to take a single Auth0 user field (email) and SAML map it to multiple fields in the assertion.
> 
> The addon documentation says
> 
>> mappings (object): The mappings between the Auth0 user profile and the output attributes on the SAML Assertion. Each "name" (left side) represents the property name on the Auth0 user profile. Each "value" (right side) is the name (including namespace) for the resulting SAML attribute in the assertion. Default mapping is shown above.
> 
> How does one map a single Auth0 field to multiple SAML fields using the Addon?
> 
> I can see that in a rule, it would be straight forward : https://auth0.com/rules/saml-attribute-mapping
> but how can I do it in the addon?

gene1wood:
Auth0 support writes

> I have checked this internally. It will not be possible to map a single Auth0 field to multiple SAML fields using the Addon. You will need to use a rule to do this mapping as you have mentioned (https://auth0.com/rules/saml-attribute-mapping).

april:
I definitely think we should learn _more_ towards using the official addon, and less on custom rules. Custom rules have a long-term overhead, and slow down every authentication.

And I disagree on `_email` as an attribute: it describes what it is, and the leading underscore indicates that it's a private attribute. This is already used as nomenclature in the user object: `_id` and `_HRData`. I don't think `_copy_of_email_used_only_for_saml_authentication` really adds that much.

dividehex:
Actions have replaced Rules and the new action API allows multiple SAML fields to be mapped to a single Auth0 user attribute.

PR #347, opened 2020-11-11
Remove account linking support for pre-deratcheting cases

gene1wood:
Now that we've manually force linked all profiles in Auth0
we no longer need to support Case 3 of account linking.

This simplifies the logic to account for just two cases.

### DIFF
--- a/rules/link-users-by-email-with-metadata.js
+++ b/rules/link-users-by-email-with-metadata.js
@@ -21,18 +21,6 @@ function (user, context, callback) {
     return callback(null, user, context);
   }
 
-  // Lower is better - used by linking ratcheting logic
-  const matchOrder = {'Mozilla-LDAP': 0,
-                      'Mozilla-LDAP-Dev': 0,
-                      'firefoxaccounts': 1,
-                      'github': 2,
-                      'google-oauth2': 3,
-                      'email': 4,
-                      'unknown': 5 // Always lowest
-                     };
-  // How old an account should be to be considered "new" to the system for linking, in seconds
-  const user_ratchet_link_delay_sec = 300;
-
   const userApiUrl = auth0.baseUrl + '/users';
   const userSearchApiUrl = auth0.baseUrl + '/users-by-email';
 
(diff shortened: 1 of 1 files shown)


rules/link-users-by-email-with-metadata.js
```diff
@@ -47,128 +35,62 @@ function (user, context, callback) {
...
-    // If the only account in auth0 database is the one they just logged in with (data.length === 0)
-    // the user is logging in with either a linked or unlinked account and there are no other accounts
-    // continue with login
-    if (data.length === 0) {
+    if (data.length === 1) {
+      // The user logged in with an identity which is the only one Auth0 knows about
+      // Do not perform any account linking
       return callback(null, user, context);
     }
+    var firstSearchResult = data[0] || undefined;
```
april, line 51: Should be `const`, I know I know, but you should do the right thing even if other code doesn't.

rules/link-users-by-email-with-metadata.js
```diff
@@ -47,128 +35,62 @@ function (user, context, callback) {
...
-      console.log("No automatic profile linking performed due to primary profile matching current profile for: " + user.user_id);
-      return callback(null, user, context);
-    }
-
-    console.log("Performing automatic profile linking: primary profile: "+primaryUser.user_id+" is now also primary for: " + user.user_id);
-
-    // Current user providers
-    const provider = user.identities[0].provider;
-    const providerUserId = user.identities[0].user_id;
+  function linkAccount(primaryUser) {
```
april, line 88: Use `const linkAccount = primaryUser => {`

rules/link-users-by-email-with-metadata.js
```diff
@@ -47,128 +35,62 @@ function (user, context, callback) {
...
+          return linkAccount(data[1]);
+        } else {
+          // Link the current user identity into the first search result identity
+          return linkAccount(firstSearchResult);
         }
       }
+    } else {
+      // data.length is > 2 which, post November 2020 when all identities were
+      // force linked manually, shouldn't be possible
+      var error_message = "Error linking account + " + user.user_id +
```
april, line 80: Use templates here.

rules/link-users-by-email-with-metadata.js
```diff
@@ -47,128 +35,62 @@ function (user, context, callback) {
...
-
-        if (matchOrder[targetConnection] < matchOrder[primaryConnection]) {
-          console.log("Found user_id that should be primary profile used for linking, according to ratcheting logic: " + targetUser.user_id);
-          primaryUser = targetUser;
+      } else {
+        // firstSearchResult is an identity with no linked secondaries
+        if (firstSearchResult.user_id === user.user_id) {
+          // Link the current user identity into the second search result identity
+          return linkAccount(data[1]);
+        } else {
```
april, line 72: This doesn't need an else.

rules/link-users-by-email-with-metadata.js
```diff
@@ -47,128 +35,62 @@ function (user, context, callback) {
...
-
-    console.log("Performing automatic profile linking: primary profile: "+primaryUser.user_id+" is now also primary for: " + user.user_id);
-
-    // Current user providers
-    const provider = user.identities[0].provider;
-    const providerUserId = user.identities[0].user_id;
+  function linkAccount(primaryUser) {
+    // Link the current logged in identity as a secondary into primaryUser as a primary
+    console.log("Linking secondary identity " + user.user_id +
+        " into primary identity " + primaryUser.user_id);
```
april, line 91: Use templates.

rules/link-users-by-email-with-metadata.js
```diff
@@ -180,22 +102,21 @@ function (user, context, callback) {
...
+        }
+        // Finally, swap user_id so that the current login process has the correct data
+        context.primaryUser = primaryUser.user_id;
+        context.primaryUserMetadata = primaryUser.user_metadata || {};
+        return callback(null, user, context)
       });
     })
     .catch(function (err) {
-      console.log("An unknown error occured while linking accounts: " + err);
+      console.log("An unknown error occurred while linking accounts: " + err);
```
april, line 118: Templates.  :)

rules/link-users-by-email-with-metadata.js
```diff
@@ -47,128 +35,62 @@ function (user, context, callback) {
...
+      // The first search result of the /users-by-email API endpoint is the linked
+      // account for that email address if there is one. It's followed by all unlinked accounts.
 
-        var targetConnection = targetUser.identities[0].connection;
-        var primaryConnection = primaryUser.identities[0].connection;
-        // If we have a new user, its connection type should not be factored in the choice, and thus is set to unknown
-        // instead
-        if (primaryUser.user_is_new !== undefined && primaryUser.user_is_new === true) {
-          primaryConnection = 'unknown';
+      if (firstSearchResult.identities && firstSearchResult.identities.length >= 1) {
```
april, line 58: I don't think it's possible for an account to not have identities?

rules/link-users-by-email-with-metadata.js
gene1wood: Ya, I was going to do that but then looked at the style used in the Auth0 rules that Auth0 writes and they used the `function` syntax. I'll change it

april: Yeah, I think that's only necessary for the top level function.

I've tested with:

`const foo = bar => { return "blah"; };`

And it seems to work great.

rules/link-users-by-email-with-metadata.js
gene1wood: Sure thing, fixed in 22f433b824bf2f12a7bcef7b237f79739582817a

rules/link-users-by-email-with-metadata.js
gene1wood: Fixed in 22f433b824bf2f12a7bcef7b237f79739582817a

rules/link-users-by-email-with-metadata.js
gene1wood: Fixed in 22f433b824bf2f12a7bcef7b237f79739582817a

rules/link-users-by-email-with-metadata.js
gene1wood: Fixed in 22f433b824bf2f12a7bcef7b237f79739582817a

rules/link-users-by-email-with-metadata.js
gene1wood: Fixed in 22f433b824bf2f12a7bcef7b237f79739582817a

rules/link-users-by-email-with-metadata.js
gene1wood: I'll leave this to be solved another day

rules/link-users-by-email-with-metadata.js
gene1wood: It seemed from the previous logic that this was a possibility, I figure I'll leave it

april (approved):

PR #351, opened 2020-11-11
Fix the account linking logic

gene1wood:
Turns out the first element returned from /users-by-email is not special and
is not the already linked account.

This simplifies the logic to just pick the other identity.

This doesn't accomodate cases where there are more than 2 identities.

### DIFF
--- a/rules/link-users-by-email-with-metadata.js
+++ b/rules/link-users-by-email-with-metadata.js
@@ -101,7 +101,8 @@ function (user, context, callback) {
     if (!("aws_logging_sns_topic_arn" in configuration) ||
         !("aws_logging_access_key_id" in configuration) ||
         !("aws_logging_secret_key" in configuration)) {
-      throw new Error("Missing Auth0 AWS SNS logging configuration values");
+      console.log("Missing Auth0 AWS SNS logging configuration values");
+      return false;
     }
 
     const SNS_TOPIC_ARN = configuration.aws_logging_sns_topic_arn;
@@ -116,9 +117,9 @@ function (user, context, callback) {
       region: 'us-west-2',
       logger: console,
     });
-    var params = {
+    const params = {
       Message: message,
-      TopicArn: SNS_TOPIC_ARN
+      TopicArn: SNS_TOPIC_ARN,
     };
     sns.publish(params, function(err, data) {
       if (err) console.log(err, err.stack); // an error occurred
--- a/rules/link-users-by-email-with-metadata.js
+++ b/rules/link-users-by-email-with-metadata.js
@@ -39,9 +39,7 @@ function (user, context, callback) {
 
     var data = JSON.parse(body);
     // Ignore non-verified users
-    data = data.filter(function(u) {
-      return u.email_verified;
-    });
+    data = data.filter(u => u.email_verified);
 
     if (data.length === 1) {
       // The user logged in with an identity which is the only one Auth0 knows about
@@ -61,6 +59,7 @@ function (user, context, callback) {
           `over 2 identities with the email address ${user.email} ` +
           data.map(x => x.user_id).join();
       console.log(error_message);
+      publishSNSMessage(`${error_message}\n\ndata : ${JSON.stringify(data)}\nuser : ${JSON.stringify(user)}`);
       return callback(new Error(error_message));
     }
   });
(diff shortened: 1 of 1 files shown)


rules/link-users-by-email-with-metadata.js
```diff
@@ -110,12 +89,41 @@ function (user, context, callback) {
...
+        return callback(null, user, context);
       });
     })
     .catch(function (err) {
       console.log("An unknown error occurred while linking accounts: " + err);
       return callback(err);
     });
-  }
+  };
+  const publishSNSMessage = message => {
```
april, line 100: I'd move this alongside `postError`.

rules/link-users-by-email-with-metadata.js
```diff
@@ -110,12 +89,41 @@ function (user, context, callback) {
...
+    if (!("aws_logging_sns_topic_arn" in configuration) ||
+        !("aws_logging_access_key_id" in configuration) ||
+        !("aws_logging_secret_key" in configuration)) {
+      throw new Error("Missing Auth0 AWS SNS logging configuration values");
+    }
+
```
april, line 106: This should return false if it's not defined, not stop rule processing.

rules/link-users-by-email-with-metadata.js
```diff
@@ -110,12 +89,41 @@ function (user, context, callback) {
...
+    const SNS_TOPIC_ARN = configuration.aws_logging_sns_topic_arn;
+    const ACCESS_KEY_ID = configuration.aws_logging_access_key_id;
```
april, line 108: Alphabetize?

rules/link-users-by-email-with-metadata.js
```diff
@@ -110,12 +89,41 @@ function (user, context, callback) {
...
+
+    let AWS = require('aws-sdk@2.5.3');
+    let sns = new AWS.SNS({
+      apiVersion: '2010-03-31',
+      accessKeyId: ACCESS_KEY_ID,
+      secretAccessKey: SECRET_KEY,
+      region: 'us-west-2',
+      logger: console,
+    });
+    var params = {
```
april, line 119: `const`.  :)

rules/link-users-by-email-with-metadata.js
```diff
@@ -110,12 +89,41 @@ function (user, context, callback) {
...
+      Message: message,
+      TopicArn: SNS_TOPIC_ARN
+    };
+    sns.publish(params, function(err, data) {
+      if (err) console.log(err, err.stack); // an error occurred
+      else     console.log(data);           // successful response
+    });
+
```
april, line 127: Should this function return whether it succeeded or not?

rules/link-users-by-email-with-metadata.js
```diff
@@ -110,12 +89,41 @@ function (user, context, callback) {
...
+    let sns = new AWS.SNS({
+      apiVersion: '2010-03-31',
+      accessKeyId: ACCESS_KEY_ID,
+      secretAccessKey: SECRET_KEY,
+      region: 'us-west-2',
+      logger: console,
+    });
+    var params = {
+      Message: message,
+      TopicArn: SNS_TOPIC_ARN
```
april, line 121: Trailing comma.  :)

rules/link-users-by-email-with-metadata.js
```diff
@@ -110,12 +89,41 @@ function (user, context, callback) {
...
     })
     .catch(function (err) {
       console.log("An unknown error occurred while linking accounts: " + err);
       return callback(err);
     });
-  }
+  };
+  const publishSNSMessage = message => {
+    if (!("aws_logging_sns_topic_arn" in configuration) ||
+        !("aws_logging_access_key_id" in configuration) ||
```
april, line 102: alphabetize here too.  :)

april (approved):

rules/link-users-by-email-with-metadata.js
gene1wood: Ok. I got it from [the other rule that we interact with AWS in](https://github.com/mozilla-iam/auth0-deploy/blob/5e78892cf4d9ddfe9071a56a42d28a005daa1770/rules/AWS-Federated-AMR.js#L14). I'll change it.

rules/link-users-by-email-with-metadata.js
gene1wood: I'll do so in a PR next week

rules/link-users-by-email-with-metadata.js
gene1wood: Good question. I think for now there's no reason to, as we wouldn't do anything with that fact.

rules/link-users-by-email-with-metadata.js
april: Yeah, I believe that rule is required to work for Auth0 to function. This one isn't.  :)

PR #352, opened 2020-11-16
Enable publishing LDAP users from Auth0 publisher

gene1wood:
Now that we're using the new LDAP publisher[1] which doesn't publish new
LDAP users but instead merely applies updates to existing CIS users, we
need the Auth0 publisher to now create users in CIS even if they are LDAP
users.

[1]: https://github.com/mozilla-iam/cis-publishers/tree/master/cis_publishers/ldap

### DIFF
--- a/rules/activate-new-users-in-CIS.js
+++ b/rules/activate-new-users-in-CIS.js
@@ -6,7 +6,7 @@ function (user, context, callback) {
   const PERSONAPI_TIMEOUT = 5000;  // milliseconds
   const PUBLISHER_NAME = 'access_provider';
   const USER_ID = context.primaryUser || user.user_id;  // linked account, or if not linked, then the user account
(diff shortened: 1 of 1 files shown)


pwnbus (approved):

gene1wood:
Some problems that go along with this change are that the following fields which were previously seeded by LDAP upon user creation are now not seeded *and* LDAP can't update them because of the [publisher-rules](https://auth.allizom.org/.well-known/mozilla-iam-publisher-rules)

```
first_name
fun_title
last_name
phone_numbers
usernames
    LDAP-posix_id
    LDAP-posix_uid
    LDAP-uuid
identities
    mozilla_ldap_id
    mozilla_ldap_primary_email
    mozilla_posix_id
```

Naruto0-sudoy (approved):

PR #355, opened 2020-12-08
Remove OIDC conformant requirement PR to production 

pwnbus:
The original intention of this was to merge the https://github.com/mozilla-iam/auth0-deploy/pull/354 (Remove OIDC conformant requirement) to production, but it appears as though a few other changes haven't been pushed out to production either.


gene1wood (approved):
Looking at the diff of what's live in Auth0 and what's present in `master` currently (which is what this PR seeks to merge into the `production` branch) using the `auth0-ci` command `./uploader_rules.py --dry-run --rules-dir ../auth0-deploy/rules` I see that the only change that will go to production is the OIDC conformant change which is what we'd want.

PR #356, opened 2020-12-15
Change default display settings for LDAP users

gene1wood:
When the task of creating new LDAP users in CIS was moved from the LDAP publisher to this Auth0 publisher, the default display settings were changed to the overly restrictive "private" for some fields. This commit changes this back to the default display of staff


gene1wood:
This PR is deployed manually in prod already.

pwnbus (approved):

PR #358, opened 2021-01-26
Add custom mapping rules for Acoustic Production

The-smooth-operator:
This PR adds a custom mapping for the Acoustic Production SAML application.
This was done and tested for staging, so we want to go to production with it.

Running a dry run against Dev (where this will deploy to once merged to master) shows that only this new rule will take place. We should be good here.


limed (approved):

The-smooth-operator:
image

Seems that you don't have write access here @limed 
Going to fix this

The-smooth-operator:
I granted you the same level of permissions (write) I have in this account.

PR #365, opened 2021-04-19
Adding planful static idp string

rtucker-mozilla:


sciurus:
Rob says

```
i was once told that i need to make edits to the auth0-deploy code to make changes to user assertions inside auth0
so i made https://github.com/mozilla-iam/auth0-deploy/pull/365
to hopefully just add that static string of text to get passed through to planful
```

rules/planful.js
```diff
@@ -0,0 +1,15 @@
+function (user, context, callback) {
+  // Planful requires a value for each user which is always the same
+  const ALLOWED_CLIENTIDS = [
```
sciurus, line 3: @rtucker-mozilla planful's client id is missing in your latest commit.

gene1wood:
I don't think we've done this before (set a custom `app_metadata` value for an RP), but maybe Planful is different from other SAML integrations. Normally for SAML integrations, we use [SAML attribute mapping](https://auth0.com/rules/saml-attribute-mapping) to map values into the SAML assertion.

Can you link to where Planful talks about this key value pair (Searching I couldn't find anything on the web on it).

If this is a SAML attribute mapping, I'd recommend putting it in the [`SAML-configuration-mapping.js`](https://github.com/mozilla-iam/auth0-deploy/blob/master/rules/SAML-configuration-mapping.js) file instead of a standalone rule.

rules/planful.js
```diff
@@ -0,0 +1,15 @@
...
+  ];
+
+  // We only care about the above clients
+  if (!ALLOWED_CLIENTIDS.includes(context.clientID)) {
+    return callback(null, user, context);
+  }
+
+  user["IdP Entity ID"] = "fed.identropy.hostanalytics.saml2"
```
sciurus, line 11: Is this valid?

I see the user object documented at https://auth0.com/docs/rules/user-object-in-rules. I'm not sure we can just throw arbitrary keys in it.

sciurus:
Moved to https://github.com/mozilla-iam/auth0-deploy/pull/366

PR #366, opened 2021-04-20
Adding static idp entity id for planful

rtucker-mozilla:
Reworking per previous comments on pr

### DIFF
--- a/rules/SAML-configuration-mapping.js
+++ b/rules/SAML-configuration-mapping.js
@@ -5,6 +5,7 @@ function (user, context, callback) {
     'cEfnJekrSStxxxBascTjNEDAZVUPAIU2': 'stripe-subplat',  // Stripe - subplat
     'inoLoMyAEOzLX1cZOvubQpcW18pk4O1S': 'acoustic-stage',  // Acoustic stage
     'sBImsybtPPLyWlstD0SC35IwnAafE4nB': 'acoustic-prod',   // Acoustic prod
+    'H5ddlJSCfGP8ab65EnWaB2sd541CJAlM': 'planful',   // Planful prod
   };
   const client = CLIENTS[context.clientID];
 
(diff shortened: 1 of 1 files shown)


rules/SAML-configuration-mapping.js
```diff
@@ -28,6 +29,10 @@ function (user, context, callback) {
       };
 
       break;
+    case 'planful':
+      user["IdP Entity ID"] = "fed.identropy.hostanalytics.saml2"
```
sciurus, line 33: @rtucker-mozilla same question as before. Is the user something we can modify in this way? Should you maybe be adding this to the context.samlConfiguration.mappings instead?

https://auth0.com/docs/rules/context-object
https://auth0.com/docs/rules/context-object

rtucker-mozilla: @sciurus I don't have any way to know unless I can test something, and we're in a catch-22.

The line below seems to imply that I can do what I need, but again, I don't know for sure.
user.company_name = 'MOZ Corp';

duallain (commented):
Based on this document - 

https://planfulcommunity.force.com/customer/s/article/360027034051-201682596

Specifically the section about assertion requirements:

```
The information that Host Analytics requires to be passed with the assertion is as follows:

    User ID: email address of the end user (i.e. bsmith@mycompany.com)
    IdP Entity ID: See information above
```

and checking the current assertion we're providing planful, it seems like injecting an entity ID would be helpful.  Looking at our saml metadata file I think the attribute may be 'entityID' instead of 'IdP Entity ID', but in general I think this will do what we want.

gene1wood:
> I don't have any way to know unless I can test something

Much better if you just want to try stuff would be to
1. Provision the RP in `auth-dev`, our Auth0 staging environment
2. Go into the rules in dev and manually play around with setting things until you get the RP working
3. Provision the RP in `auth` and apply the tested changes you worked out in dev, into the code to be merged into prod

PR #373, opened 2021-07-07
Add SAML mapping for BitSight

gene1wood:
This adds not only the simple mapping of `email`, `given_name` and `family_name` but also
statically sets all users' BitSight roles to "Customer Portfolio Manager". This way
users can by default add new vendors to to our BitSight portfolio.

https://bugzilla.mozilla.org/show_bug.cgi?id=1718342

### DIFF
--- a/rules/SAML-configuration-mapping.js
+++ b/rules/SAML-configuration-mapping.js
@@ -5,8 +5,9 @@ function (user, context, callback) {
     'cEfnJekrSStxxxBascTjNEDAZVUPAIU2': 'stripe-subplat',  // Stripe - subplat
     'inoLoMyAEOzLX1cZOvubQpcW18pk4O1S': 'acoustic-stage',  // Acoustic stage
     'sBImsybtPPLyWlstD0SC35IwnAafE4nB': 'acoustic-prod',   // Acoustic prod
-    'H5ddlJSCfGP8ab65EnWaB2sd541CJAlM': 'planful',   // Planful prod
-    'pUmRmcBrAJEdsgRTVXIW84jZoc3wtuYO': 'planful-dev',   // Planful dev
+    'H5ddlJSCfGP8ab65EnWaB2sd541CJAlM': 'planful',         // Planful prod
+    'pUmRmcBrAJEdsgRTVXIW84jZoc3wtuYO': 'planful-dev',     // Planful dev
+    'eEAeYh6BMPfRyiSDax0tejjxkWi22zkP': 'bitsight',        // BitSight
   };
   const client = CLIENTS[context.clientID];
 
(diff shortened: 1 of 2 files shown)


rules/SAML-configuration-mapping.js
```diff
@@ -5,8 +5,9 @@ function (user, context, callback) {
     'cEfnJekrSStxxxBascTjNEDAZVUPAIU2': 'stripe-subplat',  // Stripe - subplat
     'inoLoMyAEOzLX1cZOvubQpcW18pk4O1S': 'acoustic-stage',  // Acoustic stage
     'sBImsybtPPLyWlstD0SC35IwnAafE4nB': 'acoustic-prod',   // Acoustic prod
-    'H5ddlJSCfGP8ab65EnWaB2sd541CJAlM': 'planful',   // Planful prod
-    'pUmRmcBrAJEdsgRTVXIW84jZoc3wtuYO': 'planful-dev',   // Planful dev
+    'H5ddlJSCfGP8ab65EnWaB2sd541CJAlM': 'planful',         // Planful prod
+    'pUmRmcBrAJEdsgRTVXIW84jZoc3wtuYO': 'planful-dev',     // Planful dev
```
floatingatoll, line 9: It would be slightly preferable to me for commits unrelated to BitSight to have been in a separate commit, so that if we have to revert the BitSight-specific stuff, we don't have to discard these fixes too. No need to redo anything here though.

rules/SAML-configuration-mapping.js
```diff
@@ -84,6 +91,18 @@ function (user, context, callback) {
         'lastName': 'family_name'
       };
 
+      break;
+    case 'bitsight':
+      context.samlConfiguration.mappings = {
+        'urn:oid:0.9.2342.19200300.100.1.3': 'email',
+        'urn:oid:2.5.4.3': 'given_name',
+        'urn:oid:2.5.4.4': 'family_name'
```
floatingatoll, line 99: These urn:oid entries don't appear anywhere above; are they unique to BitSight in some fashion?

floatingatoll (approved):
This seems structurally sound, and doesn't have any obvious issues that I can detect. I'm not able to review the OIDs for correctness, but any problems will be scoped to Bitsight, so that's fine.

rules/SAML-configuration-mapping.js
gene1wood: Yes, see the link in the comment below on line 102 that points to the page that talks about those values.

floatingatoll: That link isn't accessible to me without credentials, but that makes sense.

gene1wood: > That link isn't accessible to me without credentials

Oh man. That's ridiculous. What company puts their knowledge base articles behind a login. I'll contact their support to suggest they don't do this. Thanks for noticing that. Here's the page contents, I'll try to think of a better way to point to that given that the URL isn't accessible without a login.

Screenshot 2021-07-07 at 15-00-55 User Roles

gene1wood:
As we don't have a dev instance of BitSight that I can use to test, my plan is to try Monday to get approval from the IAM team to manually copy paste this rule update into Auth0 prod rule web UI to validate that this works, then back it out and merge (and continue through the deployment sequence as described in the README

gene1wood:
Atoll and I will work on this Monday morning in https://mozilla-hub.atlassian.net/browse/SE-2253

PR #378, opened 2021-08-02
Change how BitSight role mapping works

gene1wood:
Turns out that these roles are used with each log in and will change an existing user's role (as opposed to only being using on new user provisioning). As a result, we need these roles to map to user groups.

Relates to #373

### DIFF
--- a/rules/SAML-configuration-mapping.js
+++ b/rules/SAML-configuration-mapping.js
@@ -98,9 +98,21 @@ function (user, context, callback) {
         'urn:oid:2.5.4.3': 'given_name',
         'urn:oid:2.5.4.4': 'family_name'
       };
-      // Grant every user the Customer Portfolio Manager role
+      // Assign BitSight roles based on group membership.
       // https://help.bitsighttech.com/hc/en-us/articles/360008185714-User-Roles
-      user.app_metadata['bitsight_user_role'] = 'Customer Portfolio Manager';
+      // https://help.bitsighttech.com/hc/en-us/articles/231658167-SAML-Documentation
+      // Possible values :
+      //   Customer User
+      //   Customer Admin
+      //   Customer Group Admin
+      //   Customer Portfolio Manager
+      if (user.groups.includes('mozilliansorg_bitsight-admins')) {
+        user.app_metadata['bitsight_user_role'] = 'Customer Admin';
+      } else if (user.groups.includes('mozilliansorg_bitsight-users')) {
+        user.app_metadata['bitsight_user_role'] = 'Customer Portfolio Manager';
+      } else {
+        user.app_metadata['bitsight_user_role'] = 'Customer User';
+      }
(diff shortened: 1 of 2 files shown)


rules/SAML-configuration-mapping.js
```diff
@@ -98,9 +98,21 @@ function (user, context, callback) {
...
-      user.app_metadata['bitsight_user_role'] = 'Customer Portfolio Manager';
+      // https://help.bitsighttech.com/hc/en-us/articles/231658167-SAML-Documentation
+      // Possible values :
+      //   Customer User
+      //   Customer Admin
+      //   Customer Group Admin
+      //   Customer Portfolio Manager
+      if (user.groups.includes('mozilliansorg_bitsight-admins')) {
+        user.app_metadata['bitsight_user_role'] = 'Customer Admin';
+      } else if (user.groups.includes('mozilliansorg_bitsight-users')) {
```
floatingatoll, line 111: Is this as intended - users are "portfolio managers", non-members are "users"?

tests/SAML-configuration-mapping.test.js
```diff
@@ -112,3 +114,16 @@ test('BitSight', () => {
...
 });
 
+test('BitSight admins', () => {
+  _context.clientID = 'eEAeYh6BMPfRyiSDax0tejjxkWi22zkP';
+  _user.groups = ['mozilliansorg_bitsight-admins', 'mozilliansorg_bitsight-users'];
+  output = rule(_user, _context, configuration, Global);
+  expect(output.user.app_metadata.bitsight_user_role).toEqual('Customer Admin');
+});
+
+test('BitSight users', () => {
```
floatingatoll, line 124: Are all four possible outcomes tested here? The PR view only shows three to me but I may have failed to get context properly.

floatingatoll (approved):
I don't really know how to validate whether this works or breaks or syntax errors, but there isn't any obvious error I can see.

rules/SAML-configuration-mapping.js
gene1wood: Yes

tests/SAML-configuration-mapping.test.js
gene1wood: My intent is for there to be 3 possible outcomes, `Customer Admin` `Customer Portfolio Manager` and `Customer User`. Is there a fourth?

floatingatoll:
Resolved by f724fbe1d82cce58d4ed1b04a0e4efd84a276229.

gene1wood:
Odd, I'm not sure how this git workflow progressed but I guess the code is in `master` now. It looks like there's a separate commit to master that says it's a merge of this PR, but then this PR is closed not merged. Odd.

floatingatoll:
It might have been due to git signatures, we had to squash-merge to update production.

gene1wood:
This is deployed in prod and works, thanks all.

PR #379, opened 2021-08-06
Setting up initial rule for production testing of GHE

rtucker-mozilla:
1st commit for the GHE <--> Mozillians group mapping security scheme.

### DIFF
--- a/rules/GHE-Groups.js
+++ b/rules/GHE-Groups.js
@@ -1,0 +1,14 @@
+function (user, context, callback) {  
+  const applicationGroupMapping = {
+    'EnEylt4OZW6i7yCWzZmCxyCxDRp6lOY0': 'mozilliansorg_ghe_saml-test-integrations_users',
+  };
+  
+  if(applicationGroupMapping[context.clientID] !== undefined){
+    if(!user.app_metadata.groups.includes(applicationGroupMapping[context.clientID])){
+        context.redirect = {
+   				url: "https://sso.mozilla.com"
+ 				};
+    }
+  }
+  return callback(null, user, context);
+}
--- a/rules/GHE-Groups.json
+++ b/rules/GHE-Groups.json
@@ -1,0 +1,4 @@
+{
+    "enabled": true,
+    "order": 1100
+}


rtucker-mozilla:
Being this is my 1st commit to this repo, I'd appreciate confirmation from @gene1wood

cknowles-moz (approved):
This matches what we attempted in testing, allowing mozilliansorg groups to control SAML access to GHE.  Noting that actual access to GHE still requires permissions on the GHE side, so this is required, but not sufficient to get access to a SAML'd GHE org.

gene1wood (commented):
What's the reason to capture this group access control in a rule as opposed to using the `apps.yml` group access control method that we use for other RPs?

By my read here's what it would look like in `apps.yml`

```
- application:
    authorized_groups:
    - mozilliansorg_ghe_saml-test-integrations_users
    authorized_users: []
    client_id: EnEylt4OZW6i7yCWzZmCxyCxDRp6lOY0
    display: false
    logo: auth0.png
    name: GHE
    op: auth0
    url: https://auth.mozilla.auth0.com/samlp/EnEylt4OZW6i7yCWzZmCxyCxDRp6lOY0
```

rtucker-mozilla:
I didn’t see a way to do a redirect to a place of our choosing via apps.yml. We intend to redirect to a GitHub page eventually. Also this makes it a 1 liner. It’s expected there will be a ton of these.

gene1wood:
> I didn’t see a way to do a redirect to a place of our choosing via apps.yml

Ah got it, I didn't realize that redirect to the sso dashboard if the user isn't authorized was a placeholder.

> It’s expected there will be a ton of these.

Does that mean that there will be a ton of Auth0 Client/Applications, each with a different `client_id` that will map to an access group?

rtucker-mozilla:
There will be a large amount of Auth0 applications. There could also be a need to redirect to different endpoints based on the failed application, so this made sense to implement here, vs in app.yml. I absolutely could be wrong though.

rtucker-mozilla:
Is there any reason to not merge this?
After the work is completed to allow for this functionality in the app.yml file, it could be removed.

gene1wood:
>  After the work is completed to allow for this functionality in the app.yml file, it could be removed.

If the reason to create this rule is the missing functionality in `apps.yml` to redirect users who fail the stage 1 authorization check to differing URLs based on the RP, then I think it all comes down to when this capability is needed by and whether this new rule or changing the existing rule that consumes `apps.yml` will be completed in time.

Also keep in mind, the current error flow when a user fails stage 1 authorization requires not passing an arbitrary redirect URL as a parameter to the SSO Dashboard due to the security risk of an [open redirect](https://cwe.mitre.org/data/definitions/601.html)

>> It’s expected there will be a ton of these.

> Does that mean that there will be a ton of Auth0 Client/Applications, each with a different client_id that will map to an access group?

I ask this because if the current plan for GitHub Enterprise integration with Auth0 involves creating a ton of different Auth0 application/clients, we may be going down the wrong path and should consider it.

rtucker-mozilla:
I can't imagine not needing to create separate Auth0 applications as every GithHub org has a different audience and callback.

floatingatoll:
Servicedesk in the past asked us to create custom outcomes for "access denied" from Auth0, which I filed as mozilla-iam/sso-dashboard-configuration#256, because the current outcome is not sufficient to provide a good experience to staff and directs people who need access to the wrong next-hop in many cases. Now the GitHub Enterprise effort has need to do the same thing, and has advanced the idea further than simply altering the email address to instead redirecting people to a page with clear instructions on how to proceed. I encourage this rule as a 'proof of concept', and if it goes well, IAM might someday incorporate and standardize this.

gene1wood:
> I encourage this rule as a 'proof of concept' that seems safe for production use (once tested)

I'd say better would be to add this feature to the existing stage 1 authorization (in [`AccessRules.js`](https://github.com/mozilla-iam/auth0-deploy/blob/master/rules/AccessRules.js) as opposed to creating an additional rule specific to GHE such that rule order affects whether this GHE rule governs access or the `AccessRules.js` governs access.

rtucker-mozilla:
@gene1wood I'm going to merge this.

rtucker-mozilla:
I didn't go to through every rule, but here are a few others that are doing access control in them. I think there is enough prior art to support this path.

rtucker-mozilla:
https://github.com/mozilla-iam/auth0-deploy/blob/master/rules/duosecurity.js
https://github.com/mozilla-iam/auth0-deploy/blob/master/rules/SAML-Braintree-attribute.js

PR #381, opened 2021-08-12
Adding ability to restrict access to Github Enterprise Organisations …

rtucker-mozilla:
I believe this is the way you would like it @gene1wood .

### DIFF
--- a/rules/AccessRules.js
+++ b/rules/AccessRules.js
@@ -273,3 +273,20 @@ function (user, context, callback) {
     return get_verified_access_rules(access_decision, fake_access_file_conf);
   });
 }
+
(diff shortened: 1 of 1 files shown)


gene1wood:
@rtucker-mozilla I don't think a given single Auth0 rule can contain multiple functions within it like this (we don't have any at least). I'd meant to add the functionality to the existing function. But lemme do this, I'll back away and you and your team solve this as you see fit, I think I'm mucking up the gears. I think you get what I'm suggesting, factor that in in whatever way looks best.

rtucker-mozilla:
The existing function isn't something that I would want to mess with. It's very long and complicated.

PR #384, opened 2021-09-15
Enable debugging the federated AWS AMR ID Token field in prod

gene1wood:


floatingatoll (approved):

floatingatoll:
With this patch, we are able to diagnose "AMR too large" issues that are linked to this comment in the Auth0 GHE rules:

https://github.com/mozilla-iam/auth0-deploy/blob/master/rules/AWS-Federated-AMR.js#L64

duallain (commented):
Adding the testrp makes a ton of sense.  It feels like maybe the bitsight and ghe rules are different work units?

floatingatoll:
Yes, those bitsight/ghe commits were due to PR drift since the PR was filed. Cleaned up with git cli and pushed to master, and will go out in today's deploy if not already live.

PR #385, opened 2021-09-15
Deploy GitHub IAM beta rule to production

floatingatoll:


floatingatoll:
GitHub merge wants to pull in CI, but that's (apparently) excluded from the production branch. A CLI merge doesn't, so we use CLI merge in a new PR.

PR #389, opened 2021-09-16
Fix unsigned commits in branch master

floatingatoll:
Due to an unsigned commit in the history, I need to rewrite and sign three commits in order to repair things. I have noted in each commit where I was not previously the author that I've done so, and the third commit is my own so no authorship note is included there.

This pull request **must not be merged**, but instead must be **force pushed** onto master. This is a **history rewrite** event.

The following commits will have their authorship changed as a result of this:

0fa187f (gene1wood) -> 2657641 (floatingatoll)
996c70e (rtucker-mozilla) -> f826448 (floatingatoll)

The following merge commit will be removed as it's not possible to reconstruct it with the history rewrite:

f724fbe (floatingatoll) -> [removed]

The following commit has been resigned, but I was already the author of it:

11ddfab (floatingatoll) -> b189d15 (floatingatoll)

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,10 @@ Please note that for any large change (i.e. anything but a single rule change), 
 5. [Test in prod](https://mana.mozilla.org/wiki/display/SECURITY/Auth0+manual+testing) 
    to make sure everything works and rollback if it doesn't.  
 
+##### What if the CodeBuild fails?
+
+`HTTP 409 Conflict` means that the unique ordering ID stored in a new rule's .json file is not unique across all .json files. Alter the ordering ID in the new rule's .json and try again.
+
 ## Testing
 
 Test are run by GitHub actions on every Pull Request.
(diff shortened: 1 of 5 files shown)


gene1wood:
Weird. I don't have the capability to make unsigned commits so it looks like commit 0fa187fa7bd7b4d5a24eb59da56bab1721208052 was authored by me and committed by floatingatoll 8 days later

```
Commit 0fa187fa
by Gene Wood, 08/02/2021 04:26 PM
committed by floatingatoll, 08/10/2021
```

without a signature. This likely relates to this PR that got closed and committed separately : https://github.com/mozilla-iam/auth0-deploy/pull/378#issuecomment-897915374

floatingatoll:
Yes, that's correct; the PR was an unsigned commit, which I merged using the GitHub website UI, leading to the signed merge commit but the unsigned PR commit within it. This breaks all the things, and so this is cleanup from that event.

@gene1wood If you'd like to do the honors here, you have an org admin account; otherwise I'm planning to work with Alan on it when he's available.

duallain:
I think this should no longer be needed in exactly this form.

floatingatoll:
Fixed by c776152.

PR #391, opened 2021-09-20
Now doing a check in mozillians to confirm githubUsername is set 

rtucker-mozilla:
In addition to confirming this is an application to process, now confirm that the user has a valid githubUsername and redirect to wiki page with an explanation

### DIFF
--- a/rules/GHE-Groups.js
+++ b/rules/GHE-Groups.js
@@ -7,7 +7,7 @@ function (user, context, callback) {
 
   const fetch = require('node-fetch@2.6.0');
   const AUTH0_TIMEOUT = 5000;  // milliseconds
-  const PERSONAPI_BEARER_TOKEN_REFRESH_AGE = 64800;  // 18 hours
+  const PERSONAPI_BEARER_TOKEN_REFRESH_AGE = 64770;  // 18 hours - 30 seconds for refresh timeout allowance
   const PERSONAPI_TIMEOUT = 5000;  // milliseconds
   const USER_ID = context.primaryUser || user.user_id;
   const getBearerToken = async () => {
@@ -80,6 +80,7 @@ const getPersonProfile = async () => {
             // If somehow dinopark allows a user to store an empty value
             // Let's set to null to be redirected later
             if(githubUsername.length === 0){
+              console.log("empty HACK#GITHUB")
               githubUsername = null;
             }
             //console.log("githubUsername: " +  githubUsername);
@@ -91,7 +92,7 @@ const getPersonProfile = async () => {
           // confirm the user has a githubUsername stored in mozillians, otherwise redirect
           if(!user.app_metadata.groups.includes(applicationGroupMapping[context.clientID]) || githubUsername === null){
             context.redirect = {
-               url: "https://wiki.mozilla.org/GitHub/SAML_issues"
+               url: configuration.github_enterprise_wiki_url
              };
           }
           return callback(null, user, context);
--- a/rules/GHE-Groups.js
+++ b/rules/GHE-Groups.js
@@ -1,14 +1,103 @@
-function (user, context, callback) {  
+function (user, context, callback) {
+  // dictionary of applications and their related mozillians groups to worry about
+
   const applicationGroupMapping = {
     'EnEylt4OZW6i7yCWzZmCxyCxDRp6lOY0': 'mozilliansorg_ghe_saml-test-integrations_users',
   };
-  
-  if(applicationGroupMapping[context.clientID] !== undefined){
-    if(!user.app_metadata.groups.includes(applicationGroupMapping[context.clientID])){
-        context.redirect = {
-   				url: "https://sso.mozilla.com"
- 				};
+
+  const fetch = require('node-fetch@2.6.0');
+  const AUTH0_TIMEOUT = 5000;  // milliseconds
+  const PERSONAPI_BEARER_TOKEN_REFRESH_AGE = 64800;  // 18 hours
+  const PERSONAPI_TIMEOUT = 5000;  // milliseconds
+  const USER_ID = context.primaryUser || user.user_id;
+  const getBearerToken = async () => {
+    // if we have the bearer token stored, we don't need to fetch it again
+   if (global.personapi_bearer_token &&
+        global.personapi_bearer_token_creation_time &&
+        Date.now() - global.personapi_bearer_token_creation_time < PERSONAPI_BEARER_TOKEN_REFRESH_AGE) {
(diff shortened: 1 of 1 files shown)


rules/GHE-Groups.js
```diff
@@ -1,14 +1,103 @@
...
+      // store the bearer token in the global object, so it's not constantly retrieved
+      global.personapi_bearer_token = data.access_token;
+      global.personapi_bearer_token_creation_time = Date.now();
+
+      console.log(`Successfully retrieved bearer token from Auth0`);
+      return global.personapi_bearer_token;
+    } catch (error) {
+      throw Error(`Unable to retrieve bearer token from Auth0: ${error.message}`);
+    }
+  };
```
floatingatoll, line 48: Is this brace indented wrong, maybe? Something seems off in my mental count of nesting and I wanted to make sure it's not an outright missing line.

rules/GHE-Groups.js
```diff
@@ -1,14 +1,103 @@
...
+  const fetch = require('node-fetch@2.6.0');
+  const AUTH0_TIMEOUT = 5000;  // milliseconds
+  const PERSONAPI_BEARER_TOKEN_REFRESH_AGE = 64800;  // 18 hours
+  const PERSONAPI_TIMEOUT = 5000;  // milliseconds
+  const USER_ID = context.primaryUser || user.user_id;
+  const getBearerToken = async () => {
+    // if we have the bearer token stored, we don't need to fetch it again
+   if (global.personapi_bearer_token &&
+        global.personapi_bearer_token_creation_time &&
+        Date.now() - global.personapi_bearer_token_creation_time < PERSONAPI_BEARER_TOKEN_REFRESH_AGE) {
```
floatingatoll, line 17: This rule is allowed to take up to 20 seconds to run by Auth0 before it's forcibly terminated, so it's worth making sure we have at least 20 seconds of validity left on the token, in addition to making sure it's valid at the instant that we run this. Perhaps simply `time < (REFRESH_AGE - 30)` or similar? Otherwise, users who hit the boundary condition will get hard-to-trace issues such as logins that fail midway through at a specific time of day (18h cyclic with a few minutes of drift), but work when they try again or at any other time of day.

rules/GHE-Groups.js
```diff
@@ -1,14 +1,103 @@
...
-    }
-  }
-  return callback(null, user, context);
-}
\ No newline at end of file
+        getPersonProfile().then(profile => {
+          let githubUsername = null;
+          // Get githubUsername from person api, otherwise we'll redirect
+          try {
+            githubUsername = profile.usernames.values['HACK#GITHUB'];
```
floatingatoll, line 79: Is this hard-coded value a specific keyword from CIS or something?

rules/GHE-Groups.js
```diff
@@ -1,14 +1,103 @@
...
+            // If somehow dinopark allows a user to store an empty value
+            // Let's set to null to be redirected later
+            if(githubUsername.length === 0){
+              githubUsername = null;
```
floatingatoll, line 83: You may want to log this event somehow, since it would indicate some sort of profile damage in the database. We might not ever use that log, so it's not urgent, but on the theory that we'd want to know *if* we were looking at logs, anyways.

rules/GHE-Groups.js
```diff
@@ -1,14 +1,103 @@
...
+            //console.log("githubUsername: " +  githubUsername);
+          } catch (e){
+            // Unable to do the githubUsername lookup
+          }
+          
+          // Confirm the user has the group defined from mozillians matching the application id
+          // confirm the user has a githubUsername stored in mozillians, otherwise redirect
+          if(!user.app_metadata.groups.includes(applicationGroupMapping[context.clientID]) || githubUsername === null){
+            context.redirect = {
+               url: "https://wiki.mozilla.org/GitHub/SAML_issues"
```
floatingatoll, line 94: Can this URL be stored in the `configuration` object?

floatingatoll (approved):
I'm assuming this has been tested in dev and you're able to complete an LDAP login process without impact from the rule, so in that context, r+. I left comments below that I encourage taking into account for future pushes, unless the brace-indent one is actually material and relevant.

rules/GHE-Groups.js
rtucker-mozilla: yep

rules/GHE-Groups.js
rtucker-mozilla: I resolved this by setting the lifetime to 64770

rules/GHE-Groups.js
rtucker-mozilla: it's correct

rules/GHE-Groups.js
rtucker-mozilla: yep!

rules/GHE-Groups.js
rtucker-mozilla: done!

rules/GHE-Groups.js
```diff
@@ -1,14 +1,104 @@
...
-}
\ No newline at end of file
+        getPersonProfile().then(profile => {
+          let githubUsername = null;
+          // Get githubUsername from person api, otherwise we'll redirect
+          try {
+            githubUsername = profile.usernames.values['HACK#GITHUB'];
+            // If somehow dinopark allows a user to store an empty value
+            // Let's set to null to be redirected later
+            if(githubUsername.length === 0){
```
bkochendorfer, line 82:
```suggestion
            if(githubUsername.length === 0) {
```

rules/GHE-Groups.js
```diff
@@ -1,14 +1,104 @@
...
+              console.log("empty HACK#GITHUB")
+              githubUsername = null;
+            }
+            //console.log("githubUsername: " +  githubUsername);
+          } catch (e){
```
bkochendorfer, line 87:
```suggestion
          } catch (e) {
            console.log(`Error getting GitHub username from user profile`);
```

rules/GHE-Groups.js
```diff
@@ -1,14 +1,104 @@
...
+            // Unable to do the githubUsername lookup
+          }
+          
+          // Confirm the user has the group defined from mozillians matching the application id
+          // confirm the user has a githubUsername stored in mozillians, otherwise redirect
+          if(!user.app_metadata.groups.includes(applicationGroupMapping[context.clientID]) || githubUsername === null){
```
bkochendorfer, line 93:
```suggestion
          if(!user.app_metadata.groups.includes(applicationGroupMapping[context.clientID]) || githubUsername === null) {
```

PR #398, opened 2021-09-22
sync from prod auth0 rules into git master branch

floatingatoll:
Already-live production changes that need to be incorporated back into production branch.

### DIFF
--- a/rules/GHE-Groups.json
+++ b/rules/GHE-Groups.json
@@ -1,4 +1,4 @@
 {
-    "enabled": true,
+    "enabled": false,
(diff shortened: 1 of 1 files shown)


floatingatoll:
These are already live in prod, so merging directly.

floatingatoll:
(This is since deployed, but I'm picking up the pieces from repo insanity.)

bkochendorfer (approved):

floatingatoll:
Other work has occurred that prevents this from merging. A future round of work in the repo will clean this up.

PR #402, opened 2021-09-23
import current auth0 rules from production

floatingatoll:
dumped auth0 rules from prod before proceeding with deploy


floatingatoll:
There has been a miscommunication and we're not proceeding at this time. Closing and will revisit in a future deploy.

PR #414, opened 2022-10-31
Live capture of auth0 dev rules

dividehex:
This PR applies the live auth0 dev tenant (reality) rules to the code base.  This is in preparation of doing the same capture of prod.

### DIFF
--- a/rules/GHE-Groups.js
+++ b/rules/GHE-Groups.js
@@ -1,9 +1,18 @@
 function (user, context, callback) {
+
   // dictionary of applications and their related mozillians groups to worry about
 
+  // THIS IS THE DEV SITE AND GHE MAPPINGS FOR PROD WON'T WORK IN DEV
   const applicationGroupMapping = {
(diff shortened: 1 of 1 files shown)


rules/GHE-Groups.js
```diff
@@ -1,10 +1,19 @@
 function (user, context, callback) {
+
   // dictionary of applications and their related mozillians groups to worry about
 
+  // THIS IS THE DEV SITE AND GHE MAPPINGS FOR PROD WON'T WORK IN DEV
   const applicationGroupMapping = {
-    'EnEylt4OZW6i7yCWzZmCxyCxDRp6lOY0': 'mozilliansorg_ghe_saml-test-integrations_users',
```
floatingatoll, line 5: I think this is fine. Whatever ends up live in dev can be adjusted if needed, so long as we've done a prod->dev copy recently.

floatingatoll (approved):

PR #418, opened 2022-11-01
Missing changes in production

dividehex:


floatingatoll (approved):
I agree that the production branch should have GHE enabled. I don't know the exact history of test-rules.yaml, but it doesn't seem harmful. r+

PR #419, opened 2022-11-03
Test updates and various fixes for Node 12

dividehex:
This PR does the follow:
* Records the existing custom social connections that exist in Dev, then updates them with small fixes
* Converts all functions from anonymous to named.
* * This is recommended by auth0 and helps with debugging the stack.
* * It also allows basic Node 12 syntax checking to succeed (node --check)
* Rules that were missing test suites all together no have bare bone test suites so test coverage can be expanded in future
* * Includes a test suite example file
* We add a `auth0` object the can be expanded to mock the same that is available in the Auth0 webtask containers
* Updates the npm package.json to include some needed modules

Currently, the rules and connections are already deployed to the `dev` tenant.


rules/AWS-Federated-AMR.js
```diff
@@ -1,6 +1,6 @@
 /*jshint esversion: 6 */
 
-function (user, context, callback) {
+function AWSFederatedAWR(user, context, callback) {
```
floatingatoll, line 3: Proofreading note: AWR should be AMR.

floatingatoll (approved):
This seems fine and is live in Dev.

PR #425, opened 2022-11-07
Merge master into production

dividehex:
This PR is identical to https://github.com/mozilla-iam/auth0-deploy/pull/419.  This deployed to the Auth0 Prod tenant.


connections/bugzilla.js
```diff
@@ -0,0 +1,24 @@
+function bugzillaConnection(access_token, ctx, callback) {
+  request.get('https://bugzilla.allizom.org/api/user/profile', {
```
floatingatoll, line 2: This is a dev hostname and likely won't function correctly in prod. I think we need to fetch the prod connections and then merge the Node.js changes into them?

connections/firefoxaccounts.js
```diff
@@ -0,0 +1,72 @@
...
+  // Auth0 already verified the id_token and it's signature at this stage
+  // See docs at https://developer.mozilla.org/en-US/docs/Mozilla/Tech/Firefox_Accounts/Introduction
+  // See also https://github.com/mozilla/fxa-oauth-server/issues/519#issuecomment-367196241 for information on fields
+  // Note that we assert this email is verified by FxA, though we do not check `amr` contains `email`. This is because
+  // We currently assert `email_verified` to be true if it has ever been verified, not if it's "just been verified"
+  var jwt = require('jsonwebtoken');
+  var id_token = jwt.decode(ctx.id_token);
+
+  // Request additional profile info, such as picture, locale, etc.
+  request.get('https://profile.stage.mozaws.net/v1/profile', {
```
floatingatoll, line 28: This is a stage hostname, see prior concern.

rules/temporary-LDAP-re-reintegration.js
```diff
@@ -1,4 +1,4 @@
-function (user, context, callback) {
+function temporaryLDAPReReintergration(user, context, callback) {
```
floatingatoll, line 1: Proofreading note: I didn't notice this earlier, but 'Reintegration' is misspelled. It isn't urgent to fix now, but duly noted for later.

floatingatoll (changes requested):

connections/bugzilla.js
floatingatoll: Per conversation, this is not fully valid — the prod connectors aren't synced yet, and there's no bugzilla connector in prod — however, the entire 'connections/' directory isn't part of any deploy or CI, so this is more like WIP being pulled along so we can update prod to Node.js 12 in a timely manner. Approving review with the understanding that this will be made more consistent.

connections/firefoxaccounts.js
floatingatoll: (See prior unconcern.)

floatingatoll (approved):

dividehex:
I'll follow up with fixes for these in another PR

PR #435, opened 2023-05-03
Allow missing rcontext.request.query (eg. from CLI applications)

globau:


globau:
Turns out there's a _lot_ more that needs fixing to support refresh tokens.  As these changes would be high risk we're looking at alternatives to refresh tokens.

PR #443, opened 2023-07-25
Set additional claims in idToken to see what gets passed

gene1wood:
This takes the group list which we put in the `amr` claim and also tries to set it in the `id` and `oaud` claims, both of which AWS says they accept


floatingatoll:
I referred to this and work resulted in #444, which implements the same.

PR #444, opened 2023-07-25
set not only idToken.amr but also .oaud, .id

floatingatoll:
Auth0 seems to be ignoring writes to the .amr field, so we've added two more copies of the group data to .oaud and .id and confirmed those are passed through to endpoints successfully.


floatingatoll:
This change is already live in production as part of incident response; this PR resyncs the repo copy of this rule to match prod.

PR #445, opened 2023-07-25
Change the "id" claim to a / delimited string

gene1wood:


floatingatoll:
This change works as intended, resulting in an .id token that contains /a/aa/aaa/ for groups (a,aa,aaa). Live in prod.

floatingatoll:
Merged as 9704ccb.

PR #452, opened 2023-11-28
Make sure primary account is the LDAP account when linking

dividehex:
This PR ensures the primary account is an LDAP account when linking accounts during the Auth0 rule workflow.  If neither accounts being linked are LDAP primary, then the primary and secondary relationship doesn't matter.

### DIFF
--- a/rules/link-users-by-email-with-metadata.js
+++ b/rules/link-users-by-email-with-metadata.js
@@ -13,7 +13,7 @@
  */
 
 function linkUsersByEmailWithMetadata(user, context, callback) {
-  const fetch = require('node-fetch@2.6.1');
+  const fetch = require("node-fetch@2.6.1");
 
   // Check if email is verified, we shouldn't automatically
   // merge accounts if this is not the case.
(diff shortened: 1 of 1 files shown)


rules/link-users-by-email-with-metadata.js
```diff
@@ -25,85 +25,125 @@ function linkUsersByEmailWithMetadata(user, context, callback) {
...
+    .catch((err) => {
+      return callback(err);
+    });
+
+  const linkAccount = (otherProfile) => {
+    // LDAP takes priority being the primary identity
+    // So we need to determine if one or neither are LDAP
+    var primaryUser = {};
+    var secondaryUser = {};
+    if (user.user_id.startsWith('ad|Mozilla-LDAP')) {
```
gcoxmoz, line 82: I think you should check both `user` and `otherProfile`.  The `else` here hides "what if neither are LDAP" and "what if both are LDAP" and while that's probably not likely, finding this later is super expensive.

dividehex: First, both being LDAP should _**NOT**_ be possible (unless we add another LDAP connector) so that removes that case.  Second, if the current account primary is LDAP, then link `other` to `current.  Third, if current is NOT primary LDAP, then the _maybe_ `other` is LDAP, therefore link to that one.  In the fourth and last case, if neither are LDAP, use logic from third case because it just doesn't matter.  We would simply be linking non-ldap to non-ldap with we can also.  And in that case primary doesn't matter and we still need to link them.  There really is no need to check for all 4 case.  We really only care about A/B vs B/A.

floatingatoll (approved):
I am comfortable with seeing this go through dev/stage/prod. We've discussed extensively and I think this will address our concerns.

rules/link-users-by-email-with-metadata.js
floatingatoll: We discussed further on a call and have identified two users, one which has been broken for four years, that are impacted by this. I recommend logging if _both_ are LDAP and then proceeding with linking one to the other, and setting up an Auth0 lin[kt]ing cron to detect it if it ever occurs again.

dividehex: I would opt not link the accounts and instead allow the login to proceed with an error to the logs.  This case can be check with the linking code.  The case would be where accounts share the same "email" but not the same primary user id.

floatingatoll (approved):
I think this is reasonable; there's lint changes mixed in, so I have a higher chance of having missed something than usual.

PR #456, opened 2023-12-13
Prevent use of restarted auth sessions

dividehex:
We don't use the /continue endpoint in our current iteration of the rules.  This prevents any user from resuming a redirected session.

### DIFF
--- a/rules/Global-Function-Declarations.js
+++ b/rules/Global-Function-Declarations.js
@@ -6,7 +6,7 @@ function globalFunctionDeclaration(user, context, callback) {
   // Since we do not use the /continue endpoint let's make sure we explictly fail with an ErrorUnauthorized
   // otherwise it is possible to continue the session even after a postError redirect is set.
   if (context.protocol === "redirect-callback") {
-    return callback(new UnauthorizedError('The /continue endpoint is not allowed'));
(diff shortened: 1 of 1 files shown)


rules/Global-Function-Declarations.js
```diff
@@ -3,6 +3,12 @@ function globalFunctionDeclaration(user, context, callback) {
   // This rule MUST be at the top of the rule list (FIRST) or other rules WILL FAIL
   // with a NON RECOVERABLE error, and thus LOGIN WILL FAIL FOR USERS
 
+  // Since we do not use the /continue endpoint let's make sure we explictly fail with an ErrorUnauthorized
+  // otherwise it is possible to continue the session even after a postError redirect is set.
+  if (context.protocol === "redirect-callback") {
+    return callback(new UnauthorizedError('The /continue endpoint is not allowed'));
```
gcoxmoz, line 9: `callback()` down at the bottom has 3 parameters; should this have `, user, context` also?
https://auth0.com/docs/rules-best-practices/rules-execution-best-practices shows an error case passing all 3 params.

dividehex: It works either way but I'll add it in here.

PR #458, opened 2024-01-29
Add mozilliansorg_http-observatory-rds group for awsSaml

dividehex:
See bug 1860024


floatingatoll (approved):
Group name matches mozilla-iam/aws-iam-identity-center-management#4.

PR #459, opened 2024-02-09
Fix the mozilla logo image link in the passwordless email

dividehex:
The email that is sent out when a user logs in with passwordless email has been broken for quite some time.  This fixes the issue.  This has already been hot patched to the auth0 prod tenant.


gcoxmoz (approved):
Old link verified busted, new link verified to exist.  Yeah let's do that.

PR #460, opened 2024-02-12
Vectra SAML modifications

dividehex:
This PR fixes the SAML for the vectra SAML application.


floatingatoll (approved):
Structurally valid rule. I get the impression this has been tested already, command it seems plausibly valid given the available data.

dividehex:
Yes, this has been tested and is running on prod already.

PR #463, opened 2024-05-22
Update GHE mapping to match current rule in production

dividehex:
GHE orgs were added via UI.  This brings those changes here to match Production.


floatingatoll (approved):
Syncs prod with repo, r+.

PR #478, opened 2024-10-23
Actions!

dividehex:
Sorry.


.gitignore
```diff
@@ -310,13 +310,49 @@ poetry.toml
...
+
+# Crash log files
+crash.log
+crash.*.log
+
+# Exclude all .tfvars files, which are likely to contain sensitive data, such as
+# password, private keys, and other secrets. These should not be part of version
+# control as they are data points which are potentially sensitive and subject
+# to change depending on the environment.
+*.tfvars
```
bheesham, line 329: Hm, do we have any standards around naming tfvars which explicitly contain secrets (e.g. `secrets.auto.tfvars`) ?

If we do, maybe we can be specific when ignoring these; otherwise yeah, a wide ignore would be best.

terraform/actions.tf
```diff
@@ -0,0 +1,443 @@
...
+      id           = auth0_action.OIDCConformanceWorkaround.id
+      display_name = auth0_action.OIDCConformanceWorkaround.name
+    }
+    configurationDumper = {
+      id           = auth0_action.configurationDumper.id
+      display_name = auth0_action.configurationDumper.name
+    }
+  }
+
+  action_flow = terraform.workspace == "prod" ? var.action_flow_prod : var.action_flow_dev
```
bheesham, line 56: Hm, I guess we swap credentials _somewhere_ to be able to apply this to the dev account? (If so, I think maybe we'd be better served with modules + a `prod/` and `dev/`, esp. considering the state bucket is currently fixed).

Or is this to have a pre-prod configuration where we can test? Would the lifecycle of a change would be:

* dev (workspace: some branch),
* dev (workspace "prod"),
* prod (workspace: some branch),
* prod (workspace: "prod")?

terraform/actions/awsSaml.js
```diff
@@ -0,0 +1,312 @@
...
+  // Only continue on auth0 prod tenant
+  if (event.tenant.id !== "auth") {
+    return;
+  }
+
+  var paramObj = {};
+
+  const clientID = event.client.client_id || "";
+  switch (clientID) {
+    case "JR8HkyiM2i00ma2d1X2xfgdbEHzEYZbS":
```
bheesham, line 15: Nonblocking nitpick: Potentially we should create a map to easier reference these `clientID`s, that way this code becomes slightly easier to reason about (and removes the need for the comment telling us which account).

Something along the lines of https://github.com/mozilla-iam/auth0-deploy/blob/1b2b75a75592fd4377fb2e4c2aaf44dd3578ab48/terraform/actions/gheGroups.js#L8-L19

.gitignore
dividehex: We don't use tfvars for secrets in this case, instead secrets are read from GCP Secrets Manager via a terraform data source.

dividehex: Either way we would not store tfvars in this public repo.

bheesham: Ah yeah fair enough.

terraform/actions/awsSaml.js
bheesham: Going to punt to the future.

bheesham (approved):

PR #483, opened 2024-11-07
Big clean up of decommed rules

dividehex:


README.md
```diff
@@ -1,121 +1,31 @@
...
+Actions are deployed with Terraform
 
+To deploy to the Auth0 dev tenant
+Make sure `AUTH0_DOMAIN`, `AUTH0_CLIENT_ID`, and `AUTH0_CLIENT_SECRET` are set for the dev tenant.
 ```
-$ python3 -m venv venv
-$ source venv/bin/activate
-$ pip install -r requirements.txt
-$ uploader_rules.py <args>
+$ cd tf
```
bheesham, line 18:
```suggestion
$ cd terraform
```

As a side-note: if we want to encourage copy/pasting, I'd suggest removing the `$ ` prefix.

bheesham (approved):
Solid. 🎉

README.md
dividehex: This PR renames the `terraform` dir to `tf` so the change here is accurate.

dividehex: I agree with removing the `$` prefix but we can do that in future PRs

bheesham: ah, whoops, missed that rename.

PR #489, opened 2024-12-09
fix Await JSON response from fetch

bheesham:
This caused a bug lower down where we'd check

        if (Object.keys(profile).length !== 0) {

The JSON promise certainly has keys, but it's not quite what we want here.


dividehex (approved):
🚀 Good fix! Approved

PR #496, opened 2025-03-14
Use prettier in CI

bheesham:
This sets up prettier and a CI job to ensure we keep a consistent style going forward.

Jira: [IAM-950](https://mozilla-hub.atlassian.net/browse/IAM-950)

Followed up by: https://github.com/mozilla-iam/auth0-deploy/pull/497


dividehex (approved):
🤔 Approved

Overall, I love prettier and I think it should be a requirement for most of our code.  My only suggestion is that instead of calling this via npm and gha, it should be wrapped in pre-commit (https://prettier.io/docs/precommit/).  Since linting and code formatting is something that should always be tested and fixed on a devs local environment before committing.

In addition to that, I think the auth0-deploy code should be migrated to `mozilla-iam/iam` where `pre-commit`<sub>1</sub> and the `pre-commit gha`<sub>2</sub> is already established.

1. https://github.com/mozilla-iam/iam/blob/main/.pre-commit-config.yaml
2. https://github.com/mozilla-iam/iam/blob/main/.github/workflows/pre-commit.yml

dividehex:
Just so I am clear on my review.  Go ahead and land this as it is.  The formatting fixes are great.  When we migrate to `mozilla-iam/iam`, that is when we can implement prettier via pre-commit.

PR #497, opened 2025-03-14
Prep for Node 22, address npm audit complaints

bheesham:
Auth0 recommends we move to Node 22 and has marked Node 18 (the version we're currently using) as "Not recommended".

This PR adds compatibility for Node 22 while addressing `npm audit`'s complaints.

Jira: [IAM-950](https://mozilla-hub.atlassian.net/browse/IAM-950)

Depends on: https://github.com/mozilla-iam/auth0-deploy/pull/496

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,21 +20,22 @@
   "homepage": "https://github.com/mozilla-iam/auth0-deploy#readme",
   "devDependencies": {
     "c8": "^9.0.0",
-    "eslint": "^8.54.0",
+    "eslint": "^9.22.0",
     "jest": "^29.7.0",
     "jest-fetch-mock": "^3.0.3",
-    "prettier": "2.8.8"
+    "prettier": "2.8.8",
+    "lodash": "^4.17.21"
   },
   "dependencies": {
     "auth0": "^4.9.0",
-    "aws-sdk": "^2.1445.0",
+    "aws-sdk": "^2.1646.0",
     "ipaddr.js": "^1.9.1",
     "js-yaml": "^4.1.0",
-    "jsonwebtoken": "^9.0.0",
-    "node-fetch": "^2.6.1",
-    "node-jose": "^2.2.0",
-    "require-from-string": "^2.0.2",
-    "strip-comments": "^2.0.1"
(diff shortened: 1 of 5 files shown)


tf/actions.tf
```diff
@@ -143,7 +143,7 @@ resource "auth0_action" "ensureLdapUsersUseLdap" {
 
   dependencies {
     name    = "aws-sdk"
-    version = "2.1646.0"
+    version = "2.1416.0"
```
dividehex, line 146: Why is aws-sdk being downgraded?

package.json
```diff
@@ -18,21 +20,22 @@
...
-    "node-jose": "^2.2.0",
-    "require-from-string": "^2.0.2",
-    "strip-comments": "^2.0.1"
+    "js-yaml": "^3.3.1",
+    "jsonwebtoken": "^9.0.2",
+    "node-fetch": "^2.7.0",
+    "lodash": "^4.17.21"
+  },
+  "engines": {
+    "node": ">=18.0.0 <v23.0.0"
```
dividehex, line 38: That doesn't look like a valid.  Maybe remove the `v`?

dividehex (commented):
++ to all the lint fixes.  It looks like this over laps with #496 which is fine.  See comments.

tf/actions.tf
bheesham: Bringing into the public: @dividehex and I had a quick chat about this.

I misunderstood how dependencies worked for Auth0 Actions. Actions may use any dependency from npm.

I used [Can I require?](https://auth0-extensions.github.io/canirequire/), but it doesn't apply to Actions, only their old system which we no longer use.

bheesham: I'll be reverting these changes in a bit -- though, still looking to sync dependencies between what we configure and what we develop with.

package.json
bheesham: Solid catch!

tf/actions.tf
bheesham: Done!

dividehex (approved):
🌺 Approved

PR #502, opened 2025-04-07
Allow for some groups to access all GitHub applications

bheesham:
Members of the infrastructure org, who admin GitHub Enterprise; and the security managers team, who do security audits, may require broad access to GitHub infra.

This commit allows for inviting them to their respective groups (`ghe_admins` and `ghe_security-managers`), which will grant them access to any GitHub applications we have defined. Instead of accepting `2 x <GitHub Organization>` invites they'll only need to accept `1 + <GitHub Organization>` invites, where that `1 +` happens when they're onboarded.

Jira: [IAM-1021](https://mozilla-hub.atlassian.net/browse/IAM-1021)


gcoxmoz (approved):
Looks like it does what it says on the tin, discussed some quibble points in Slack but nothing worth blocking over.

PR #504, opened 2025-04-14
Deny email registration for Matrix

bheesham:
We already allowed email registrations to Matrix, so we can't disable this
connection, since that would break logging in for existing users.

That leaves us with only one option: deny at pre-user-registration.

Jira: [IAM-1617](https://mozilla-hub.atlassian.net/browse/IAM-1617)

---

<details>
<summary>Plan for <code>prod</code></summary>

```
  # auth0_action.deny_registration will be created
  + resource "auth0_action" "deny_registration" {
      + code       = <<-EOT
            // Reject users from registering for an application (by client id) using a
            // specific connection.
            //
            // This is a workaround for disabling a connection entirely for an application,
            // since we may have allowed registrations already.
            //
            // If we instead disabled the connection then we'd break logins for users who
            // only have that connection available.
            //
            // DEBT(bhee): LDAP's connection name is
            // * `Mozilla-LDAP` on prod;
            // * `Mozilla-LDAP-Dev` on dev.
            //
            // If we need to deny registrations on those, for some reason, we'll need to
            // think of a better way. Connection Ids are not stable across tenants either.
            
            exports.onExecutePreUserRegistration = async (event, api) => {
              const CLIENT_CONNECTIONS_DENYLIST = {
                // Matrix, IAM-1617
                pFf6sBIfp4n3Wcs3F9Q7a9ry8MTrbi2F: ["email"],
              };
            
              const denylist = CLIENT_CONNECTIONS_DENYLIST[event.client.client_id] ?? [];
            
              if (denylist.includes(event.connection.name)) {
                return api.access.deny(
                  `Not allowed to register for ${event.client.name} using ${event.connection.name}.`
                );
              }
            
              return;
            };
        EOT
      + deploy     = true
      + id         = (known after apply)
      + name       = "denyRegistration"
      + runtime    = "node22"
      + version_id = (known after apply)

      + supported_triggers {
          + id      = "pre-user-registration"
          + version = "v2"
        }
    }

  # auth0_trigger_actions.pre_user_registration_flow will be created
  + resource "auth0_trigger_actions" "pre_user_registration_flow" {
      + id      = (known after apply)
      + trigger = "pre-user-registration"

      + actions {
          + display_name = "denyRegistration"
          + id           = (known after apply)
        }
    }

Plan: 2 to add, 0 to change, 0 to destroy.
```
</details>

### DIFF
--- a/tf/.terraform.lock.hcl
+++ b/tf/.terraform.lock.hcl
@@ -2,42 +2,48 @@
 # Manual edits may be lost in future updates.
 
 provider "registry.terraform.io/auth0/auth0" {
-  version     = "1.2.0"
-  constraints = "1.2.0"
+  version     = "1.15.0"
+  constraints = "1.15.0"
   hashes = [
-    "h1:BHqrfc+2tLSsDqh3vgHijeZVzZKs225zF9NaJo/mduE=",
-    "zh:0b80c8c7b671911a3a7672b16b0febf61658276cb881008ed2dd05da5ed31c57",
-    "zh:0d3fc8f2e767bf128ae8cabbd31cd3fe80286443a2687e289984df7f65ff2bb2",
-    "zh:1d507c15389c08468dc161a6e009387d3f56656dd9a152bcdec4d72b421aedf0",
-    "zh:22ea736c26b2f13f7f9a428e4e4765c90aa7b7f78ffe987175d91a0412b5e7ed",
-    "zh:28b74fee87ae70a29a45ee8881b966049445b487ca9547585c16f4abe7e4cc80",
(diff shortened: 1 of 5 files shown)


bheesham:
Following-up via Slack with @denschub about if we need to notify anyone, since this would be a breaking change.

denschub (approved):
I don't think we need to do any proactive comms here. We aren't breaking existing users, only new registrations, so I don't think anyone cares. I'll edit the wiki page and keep an eye on our feedback channels, but besides that, I don't see a need to be vocal about this. If I see a lot of complaints, we can put something out.

PR #520, opened 2026-02-24
Deny with the error _message_ not the error _object_

bheesham:
Turns out, we've been denying with the wrong parameter this whole time! What we're really after is `err.message`.

Jira: [IAM-1761](https://mozilla-hub.atlassian.net/browse/IAM-1761)


gcoxmoz (approved):
Reporting messages, can't be mad at that.

PR #521, opened 2026-02-24
linkUserByEmail: Split out helper functions, reduce scope of try/catches

bheesham:
The general approach here was to:

1. Move functions outside of the main exported function, `onExecutePostLogin`;
2. Reduce the size of our `try`/`catch` blocks to be more specific;
3. Add a retry, since there may be transient network errors we're failing on by only relying on the SDK's retry behaviour.

Jira: [IAM-1761](https://mozilla-hub.atlassian.net/browse/IAM-1761)

### DIFF
--- a/tf/actions/linkUserByEmail.js
+++ b/tf/actions/linkUserByEmail.js
@@ -111,15 +111,14 @@ async function linkAccount(api, mgmtClient, originalProfile, otherProfile) {
         user_id: secondaryUser.identities[0].user_id,
       }
     );
-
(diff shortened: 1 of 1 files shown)


gcoxmoz (approved):
Not my primary language, but breakup of functions makes sense to me in each of the commits.

bheesham:
Turns out, we didn't quite need this.

bheesham:
(Still useful, but maybe we'll get to it later.)

PR #528, opened 2026-04-01
Don't require MFA when using refresh tokens

bheesham:
We'd always enable MFA, which breaks the refresh token flow. To mint a refresh token, the user would have already completed a MFA challenge.

By not requiring MFA to use a refresh token we fix a long-standing issue with Lando.

Jira: [IAM-1935](https://mozilla-hub.atlassian.net/browse/IAM-1935)


gcoxmoz (approved):
Discussed some nuances of "does this open a spoof path?" and we believe not, so, this all makes sense to me.

bheesham:
Tested on dev. I've messaged folks on the Lando team to see what they'd prefer.

bheesham:
I'll merge+deploy so the Lando team can start working on implementing refresh token support.

PR #540, opened 2026-06-08
dev: use biome as an LSP

bheesham:
Configuring your editor is left as an exercise for the developer.

Jira: IAM-1761


mise.toml
```diff
@@ -1,2 +1,7 @@
 [tools]
 node = "18"
+"npm:@biomejs/biome" = "2.4.16"
```
bheesham, line 3: Needs to be rebased if https://github.com/mozilla-iam/auth0-deploy/pull/541 is merged.

bheesham:
(Typo in the branch name.)

bheesham:
Actually, use `mise use --env local npm:@biomejs/biome@2.4.16` (or something).

PR #541, opened 2026-06-08
Use Node 22

bheesham:
Auth0 no longer recommends using Node 18.

Jira: [IAM-1990](https://mozilla-hub.atlassian.net/browse/IAM-1990)


gcoxmoz (approved):
Lines up with what Auth0 wants to happen, so I like it enough to say "let's find out!"

bheesham:
DEPLOYED!

PR #545, opened 2026-07-06
Support high assurance level

bheesham:
Jira: [IAM-1989](https://mozilla-hub.atlassian.net/browse/IAM-1989)

---

tl;dr: for some RPs we need to support a stricter policy for MFA. This PR adds support for `HIGH`.

### DIFF
--- a/tf/actions/accessRules.js
+++ b/tf/actions/accessRules.js
@@ -61,6 +61,33 @@ const assuranceMax = (left, right) => {
   return right;
 };
 
+const duoConfigForLevel = (level, secrets, email) => {
+  let providerOptions;
+  let allowRememberBrowser = false;
+  if (level === "HIGH") {
+    console.info("using Duo HIGH client");
+    providerOptions = {
+      host: secrets.duo_apihost_mozilla_high,
+      ikey: secrets.duo_ikey_mozilla_high,
+      skey: secrets.duo_skey_mozilla_high,
+      username: email,
+    };
+  } else {
+    console.info("using Duo regular client");
+    providerOptions = {
+      host: secrets.duo_apihost_mozilla_medium,
+      ikey: secrets.duo_ikey_mozilla_medium,
+      skey: secrets.duo_skey_mozilla_medium,
+      username: email,
+    };
+    allowRememberBrowser = true;
+  }
+  return {
+    providerOptions,
+    allowRememberBrowser,
+  };
+};
+
 exports.onExecutePostLogin = async (event, api) => {
   console.log("Running actions:", "accessRules");
 
@@ -160,13 +187,6 @@ exports.onExecutePostLogin = async (event, api) => {
     "moc-sso-monitoring@mozilla.com", // MOC see: https://bugzilla.mozilla.org/show_bug.cgi?id=1423903
     "shared-deng-playstore@mozilla.com", // See: https://mozilla-hub.atlassian.net/browse/IAM-1938
   ];
-
-  const duoConfig = {
-    host: event.secrets.duo_apihost_mozilla,
-    ikey: event.secrets.duo_ikey_mozilla,
-    skey: event.secrets.duo_skey_mozilla,
-    username: event.user.email,
-  };
 
   // Check if array A has any occurrence from array B
   const hasCommonElements = (A, B) => {
@@ -490,10 +510,12 @@ exports.onExecutePostLogin = async (event, api) => {
 
     if (decision.granted) {
       if (decision.enableDuo && !isRefreshTokenFlow) {
-        api.multifactor.enable("duo", {
-          providerOptions: duoConfig,
-          allowRememberBrowser: true,
-        });
+        const duoConfig = duoConfigForLevel(
+          decision.aal,
+          event.secrets,
+          event.user.email
+        );
+        api.multifactor.enable("duo", duoConfig);
       }
       // Set groups, AAI, and AAL claims in idToken
       api.idToken.setCustomClaim(`${namespace}/AAI`, decision.aai);
--- a/tf/actions/accessRules.js
+++ b/tf/actions/accessRules.js
@@ -3,6 +3,64 @@ const YAML = require("js-yaml");
 const jwt = require("jsonwebtoken");
 const AWS = require("aws-sdk");
 
+// Assurance levels to accepted assurance indicators map.
+// The ordering of keys matters here.
+//
+// The entries themselves don't super matter too much, and they're going to be
+// connection-dependent. Example: for Google we seem to use
+// `HIGH_ASSURANCE_IDP`, but it's not really clear what that actually provides.
+//
+// The main thing here is that, for things we _know_ about (2FA, HARDWARE),
+// we're able to guarantee some properties of an authenticator, because for
+// Mozillians, hooked up to Duo.
+const AAI_MAPPING = {
+  // No authenticator required.
+  LOW: new Set(),
+  // Second factor required. Examples:
+  // * Google Authenticator (TOTP)
+  // * RP's authenticator settings (Google)
+  MEDIUM: new Set(["2FA", "HARDWARE", "HIGH_ASSURANCE_IDP"]),
+  // Hardware-protected authenticator required. Examples:
+  // * Platform authenticator (e.g. Windows Hello, Apple ID, Passkey)
+  // * Roaming authenticator (e.g. synced Passkey)
+  // * Hardware tokens (e.g. Yubikey)
+  HIGH: new Set(["HARDWARE"]),
+  MAXIMUM: new Set(["MAXIMUM_NOT_IMPLEMENTED"]),
+};
+
+// To compare which assurance level is higher/lower, we map the keys to a dict.
+// The order the levels are defined in matters.
+const AAI_AS_INT = Object.fromEntries(
+  Object.entries(Object.keys(AAI_MAPPING)).map(([idx, name]) => [
+    name,
+    Number(idx),
+  ])
+);
+
+// Default app requested aal to MEDIUM for all apps which do not have this set
+// in access file.
+const AAL_DEFAULT = "MEDIUM";
+
+// Given two assurance levels (e.g. `HIGH` and `LOW`), return the higher of the
+// two.
+const assuranceMax = (left, right) => {
+  const leftNum = AAI_AS_INT[left];
+  const rightNum = AAI_AS_INT[right];
+  if (leftNum === undefined && rightNum === undefined) {
+    return undefined;
+  }
+  if (leftNum === undefined && rightNum !== undefined) {
+    return right;
+  }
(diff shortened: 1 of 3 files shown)


tf/actions/accessRules.js
```diff
@@ -366,71 +426,41 @@ exports.onExecutePostLogin = async (event, api) => {
...
-        }
-      }
+    // 2 Check if user.aal is allowed for this RP
+    if (AAI_MAPPING[aal].size === 0) {
+      console.log(
+        "No required indicator in aai_mapping for this RP (mapping empty for this AAL), access will be granted"
+      );
+      aai_pass = true;
+    } else if (AAI_MAPPING[aal].intersection(aai).size > 0) {
+      const aaiPretty = Array.from(aai.values()).join(", ");
```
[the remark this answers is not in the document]
bheesham: Er... what are you smoking?

```
; node
Welcome to Node.js v22.22.3.
Type ".help" for more information.
> var haystack = new Set([1, 2, 3, 4, 5])
undefined
> var needles = new Set([1, 5])
undefined
> haystack.intersection(needles)
Set(2) { 1, 5 }
```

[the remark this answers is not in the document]
bheesham: `AAI_MAPPING[aal]` being undefined cannot be true on this line because we'd have already denied access.

tf/actions/accessRules.js
```diff
@@ -3,6 +3,81 @@ const YAML = require("js-yaml");
...
+      username: email,
+    };
+  }
+  return {
+    host: secrets.duo_apihost_mozilla_high,
+    ikey: secrets.duo_ikey_mozilla_high,
+    skey: secrets.duo_skey_mozilla_high,
+    username: email,
+  };
+};
```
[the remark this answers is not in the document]
bheesham: Yep, whoops.

kgowrabathuni (approved):

bheesham:
Tested successfully on dev:

Screenshot 2026-07-24 at 15 54 53

tf/actions/accessRules.js
```diff
@@ -41,6 +41,9 @@ const AAI_AS_INT = Object.fromEntries(
 // in access file.
 const AAL_DEFAULT = "MEDIUM";
 
+// The minimum AAL employees may have.
```
gcoxmoz: I don't love it.  If landed, I hope you'll add notes to `apps.yml` that explain that `AAL: LOW` has a new path / doesn't mean what is written.  But that's a nitpick, whatevs.

Really, I worry that there might be RPs in here that will burp because of a new minimum.  Just top of my head, there appears to be a conflation of `AD_MIN` being taken to mean 'employee' when it's LDAP users, which may have non-employees.

This might be fine!  But I am first-blush skeptical.

bheesham: Good catch, there was a bug here. The bit of code where this is used should have instead lived in a different if statement: where we check if Duo is included in their profile.

The original one didn't hit the `duo` branch, and I missed that in this implementation. The test now passes here, and [continues to pass on master](https://github.com/bheesham/auth0-deploy/actions/runs/30605332336/job/91076293016#step:6:241).

The behaviour for LDAPers hasn't changed (anymore!), and I added to the comment for `AAL_AD_MIN`. But yeah, because this was always the behaviour, the docs in `apps.yml` are still out of date. PR incoming.

bheesham: [Potentially clarified in apps.yml?](https://github.com/mozilla-iam/sso-dashboard-configuration/pull/1092)

tf/actions/accessRules.js
```diff
@@ -366,71 +416,41 @@ exports.onExecutePostLogin = async (event, api) => {
...
-            );
-            aai_pass = true;
-            break;
-          }
-        }
-      }
+    // 2 Check if user.aal is allowed for this RP
+    if (AAI_MAPPING[aal].size === 0) {
+      console.log(
+        "No required indicator in aai_mapping for this RP (mapping empty for this AAL), access will be granted"
```
gcoxmoz, line 439: I don't think we should default-allow here.  `aal` goes through checks above, and should arrive here as a valid value because of `AAL_DEFAULT` existing.  And if it doesn't, then we've missed something above, big time.

IMO, default-deny here.  It probably isn't a big win, I get it.

tf/actions/accessRules.js
```diff
@@ -304,17 +353,17 @@ exports.onExecutePostLogin = async (event, api) => {
     // depending on the connection type
 
     // Ensure all users have some AAI and AAL attributes, even if its empty
-    let aai = [];
-    let aal = "UNKNOWN";
+    const aai = new Set();
```
gcoxmoz, line 356: I could be very wrong here, but Set() being const and then edited below.. isn't that bad?

tf/actions/accessRules.js
```diff
@@ -41,6 +41,21 @@ const AAI_AS_INT = Object.fromEntries(
...
+// The minimum AAL folks coming from LDAP _may_ have.
+//
+// LDAP related logic:
+//
+// * if they're comming from LDAP, then we enable Duo;
+// * if they have Duo in their profile, then we set what kinds of factors Duo
+//   supports. (And because that means either 2FA or Hardware, the minimum is
+//   2FA).
+//
+// Before IAM-1989, you can see this in play:
```
bheesham, line 53: (er, before _and_ after. nothing related to this is changing.)

tf/actions/accessRules.js
bheesham: `const` in JS is more like "cannot be _reassigned_" and says nothing about _mutation_, which is different than other langs. e.g.:

```
; node
Welcome to Node.js v22.22.3.
Type ".help" for more information.
> const foo = 1;
undefined
> foo += 1;
Uncaught TypeError: Assignment to constant variable.
```

Buuut, because we don't need to _reassign_ `foo` when it's a `Set` to mutate it, the way I use it in code is allowed (er, I think encouraged actually). I guess this doesn't cover what happens when reallocations occur and the backing memory is reassigned.

`let`, I think, is probably along the lines of what you were expecting. Though I never _reassign_ `aai`, so the JS linter gets angry.

e.g.

```
; node
Welcome to Node.js v22.22.3.
Type ".help" for more information.
> let foo = 1;
undefined
> foo = 2;
2
> foo += 1;
3
```

`var` might also be fine, but its scoping works differently, and is confusing enough when debugging that it's usually better to avoid. (I wonder what the JS linter would have said.)

bheesham:
I chatted with @VegasNative about this, and I'm way off base! AAL is defined [here](https://pages.nist.gov/800-63-3-Implementation-Resources/63B/AAL/), and I'll quote for convenience.

> AAL1 permits single-factor authentication using a wide variety of authenticators listed in [SP 800-63B Section 4.1.1](https://pages.nist.gov/800-63-3/sp800-63b.html#411-permitted-authenticator-types). By far the most common authenticator at AAL1 is the memorized secret, but from the standpoint of meeting AAL1 requirements it is equally acceptable to use a physical authenticator such as an OTP device.

This is our `LOWEST`.

> AAL2 requires the use of two authentication factors, either (1) a physical authenticator and a memorized secret, or (2) a physical authenticator and a biometric that has been associated with it. Multi-factor authentication can be performed using either a multi-factor authenticator or through the use of two independent authenticators.

This is our `MEDIUM`.

This PR adds `HIGH`, but I don't think that's the intention here. We want to restrict which factors are used in the authentication process, not necessarily give a higher level of assurance.

Some of the logic here can be kept, but I think anything related to us saying `HIGH` AAL should be removed. I think what we really want is to add a new key to the `apps.yml` spec, to specify which authenticators should be allowed, and add that to the `AAI` ([ref for future dev](https://www.iana.org/assignments/authentication-method-reference-values)). (Maybe?)

I have a meeting scheduled with @gene1wood tomorrow to chat more.

bheesham:
(Dunno though! I'll keep reading.)

bheesham:
More info! I chatted with @gene1wood!

re: changing remembered devices in Duo: no policy; it was easier to remember in Auth0 than in Duo. I believe we should be able to change the settings _for only the SSO_ application in Duo. That would still remember clients for 30 days. It does come at the cost of an extra HTTP call, but, it's probably fine.

re: AAL The `AAL` stuff is really a mapping Mozilla's [risk standard levels](https://infosec.mozilla.org/guidelines/risk/standard_levels) (referenced in, the archived, [mozilla-iam](https://github.com/mozilla-iam/mozilla-iam#authentication-assurance) repository). I'll have to ensure the docs in mozilla-iam are copied over to Confluence, or some easier to find place.

<details><summary>A brief detour into AALs</summary>

Quoting the docs:

> The AALs are ranked using the [Mozilla Standard Levels](https://infosec.mozilla.org/guidelines/risk/standard_levels), i.e. LOW/MEDIUM/HIGH/MAXIMUM. The AAL is used to determine if a user should be granted access per relying party.

And, quoting a bit from [SP 800-63B Section 4.1.1](https://pages.nist.gov/800-63-3/sp800-63b.html#411-permitted-authenticator-types):

> AAL1 provides some assurance that the claimant controls an authenticator bound to the subscriber’s account. AAL1 requires either single-factor or multi-factor authentication using a wide range of available authentication technologies. Successful authentication requires that the claimant prove possession and control of the authenticator through a secure authentication protocol.

A memorized secret covers this when we don't prompt for MFA, and roughly corresponds to our LOW risk.

> At AAL2, authentication SHALL occur by the use of either a multi-factor authenticator or a combination of two single-factor authenticators. A multi-factor authenticator requires two factors to execute a single authentication event, such as a cryptographically-secure device with an integrated biometric sensor that is required to activate the device. Authenticator requirements are specified in [Section 5](https://pages.nist.gov/800-63-3/sp800-63b.html#sec5).

AAL2 maps to roughly our MEDIUM. Since we always ask for a password, the following applies to us:

> When a combination of two single-factor authenticators is used, it SHALL include a Memorized Secret authenticator ([Section 5.1.1](https://pages.nist.gov/800-63-3/sp800-63b.html#memsecret)) and one possession-based (i.e., “something you have”) authenticator from the following list:
> 
>    * Look-Up Secret ([Section 5.1.2](https://pages.nist.gov/800-63-3/sp800-63b.html#lookupsecrets))
>    * Out-of-Band Device ([Section 5.1.3](https://pages.nist.gov/800-63-3/sp800-63b.html#out-of-band))
>    * Single-Factor OTP Device ([Section 5.1.4](https://pages.nist.gov/800-63-3/sp800-63b.html#singlefactorOTP))
>    * Single-Factor Cryptographic Software ([Section 5.1.6](https://pages.nist.gov/800-63-3/sp800-63b.html#sfcs))
>    * Single-Factor Cryptographic Device ([Section 5.1.7](https://pages.nist.gov/800-63-3/sp800-63b.html#sfcd))

But, most of that's not really relevant to the issue: an RP requiring specific factors to be used.

</details>

PR #551, opened 2026-07-30
IAM-2053: Cleanup for nss-dev

ctbfourone:


gcoxmoz (approved):
So long and thanks for all the nss

bheesham (approved):
Confirmed that the app `Qb2ZWerstBXCn5yCXQYU7vUfLuaZ1dMB` no longer exists.

Ship it!

